PR #11, opened 2019-06-17
TargetedMS: Increase length of peptidegroup.gene column

vagisha:
Gene field in some fasta files (e.g. from Flybase) contain all possible gene names that a protein products is associated with.  This can get rather long.  I am increasing it to 2000 as we discussed over email in March.


vagisha:
Conflicts with previously committed sql scripts.  Will change script version and create another pull request.

PR #13, opened 2019-06-17
Peptidegroup gene length

vagisha:


labkey-jeckels:
@vagisha I think your source branch may need to be rebased against the target feature branch. There are 12 files showing changes here, instead of the intended 3 (two .sql files and the module's Java file).

src/org/labkey/targetedms/search/peptideSearch.jsp
```diff
@@ -16,7 +16,7 @@
  */
 %>
 <%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
-<%@ page import="org.labkey.api.ms1.MS1Urls" %>
+<%@ page import="org.labkey.api.ms2.MS2Urls" %>
```
labkey-jeckels: These changes indicate that the source branch is likely not current enough.

PR #16, opened 2019-06-20
Audit log code review

rita-gwen:
Addressing code review comments of May 13 by Josh Eckels on the audit log code merge into develop.

The original e-mail thread and my responses follow.
------------------------------------------------------------------------------
On Mon, May 13, 2019 at 11:11 AM Josh Eckels <jeckels@labkey.com> wrote:
Yes, thanks for the audit work!

I just did a code review and have some suggestions. For changes to the TargetedMS module itself, Vagisha has typically sent me a patch for comments before committing. We'll be migrating this code from SVN to Git shortly where we can use a pull request to support this workflow, but can I ask you to get pre-commit reviews for the time being?

Some comments and requests:
1. Tests (thanks for including these!)
   1. Yes, if a test needs access to the database to run, it should be included in getIntegrationTests() instead of getUnitTests().
   2. The test needs to do whatever setup is required to run. It can't assume that the user has externally created a container of a specific name or GUID. See PropertyManager.TestCase for a unit test that does this. You can test out updated code by running against a fresh database - you can just update pg.properties or mssql.properties to point to a different database name and do a "gradlew pickPg" or "gradlew pickMsSql" to run against an empty DB without losing your existing DB.
   3. I just disabled these tests for the moment in trunk with revision 63655
2. Java coding conventions
   1. SkylineAuditLogParser.TestCase includes method names that start with an upper case letter.
   2. Variables names that include underscores instead of being camelCased. log_message in TargetedMSController and p_file in AuditLogMessageExpander.parseResource(), for example.
   3. Declaring variables as a collection implementation class instead of the interface. For example, I recommend declaring AuditLogTree._children to be a Map instead of HashMap.
   4. Simple Java classes that use wrapper types for values that should never be null. For example, I think AuditLogEntry._versionId should be an int instead of an Integer.
3. Misc lint reported by IntelliJ
   1. Unused imports. SkylineAuditLogParser, for example
   2. Totally unused variables. "results" local variable in AuditLogTree.getTreeSize() and "res" in UnitTestUtil.extractLogFromZip(), for example.
   3. Using == to compare String objects. This will work many times in practice, but isn't guaranteed. Use .equals() instead. AuditLogTree.equals() and SkjylineAuditLogImporter.BuildTreeTest() for example.
   4. Potential NullPointerExceptions. SkylineAuditLogImporter.verifyPreRequisites()'s use of _logFile (I suspect the first if statement should be (_logFile == null || !_logFile.exists()) for example.
4. DB schema
Columns that have a FK on them should get an index as well. targetedms.AuditLogEntry.versionId is missing one.
5. Code location
After discussions with Brendan, our intent is to make the audit log features within Panorama a premium feature. That means at least some parts of the code should be in the separate PanoramaPremium module, not in the TargetedMS module itself. Is this something you've already considered how to accomplish, or would it be useful to brainstorm together?
Sorry if some of these seem nit-picky, but they're based on what we've established as best practices over the years. Let me know if you have questions.
Thanks,
Josh

------------------------------------------
Rita Chupalov wrote:

Looks like 1.1 and 1.2 has been already addressed by Trey, thank you!
2.1 Corrected
2.2 Corrected
2.3 Corrected. I looked for HashMap and ArrayList. Don't think I used any other containers there.
2.4 As mentioned before this is intentional. Non-null version_id indicates that this log entry is the last entry in the log file for that version of the document. Formally speaking, there is one-to-many relationship between a log entry and a document version, not null implementation would require a redundant (de-normalized) link table.

3.  Hopefully I addressed most of these. I just ran code inspection and corrected as many issues of these types as possible. Let me know if I missed anything or if there is a better way to find them.

4. That's a questionable policy, from the execution plans I've seen the index is not used for this join, but it still adds IO overhead on inserts. But I added it anyway.

5.1 - see my previous e-mail. This functionality is not part of premium feature set.


resources/schemas/dbscripts/postgresql/targetedms-19.14-19.15.sql
```diff
@@ -31,6 +31,7 @@ CREATE TABLE targetedms.AuditLogEntry (
 );
 
 CREATE UNIQUE INDEX uix_auditLogEntry_document on targetedms.AuditLogEntry USING btree (documentGUID, entryHash);
+CREATE INDEX uix_auditLogEntry_version on targetedms.AuditLogEntry USING btree (versionId);
```
labkey-jeckels: A LabKey deployment won't re-run a script that it's already executed. As such, dev machines that are already upgraded to 19.15 or later won't end up creating this index.

It can be added as a new SQL script and a separate version number bump. Unfortunately, we don't have a great solution for how to manage multiple active feature branches with schema updates, and this PR also has upgrades:

https://github.com/LabKey/targetedms/pull/4

Can you update this branch so that this index creation happens in a version update from 19.191-19.192? We will get the other PR merged first (which will get the version to 19.191) and then merge this one.

rita-gwen:
The schema files are renamed. Thanks for the review.
I don't think indexes are needed to maintain FK integrity though they will speed up the validation.

labkey-jeckels:
Agreed, the DB will maintain the FK integrity without the index, it's just for perf.

You'll also need to update TargetedMSModule.getVersion() to return 19.192. We can then merge this after the other PR has been merged.

rita-gwen:
Module version updated.

rita-gwen:
Thank you, Josh!

PR #17, opened 2019-06-25
Issue 35987: Update default guideset calculation

labkey-martyp:
Default guideset calculations include values across entire plot if no guidesets are defined, otherwise only include values to the left of the first guideset from the left.


labkey-jeckels:
@labkey-ankurj Can you check this out? Is anything else required server-side for email notifications or similar?

PR #19, opened 2019-06-28
Issue 37857: Update and add new columns to targetedms.MeasuredDriftTime

vagisha:
Update targetedms.MeasuredDriftTime to be consistent with the current Skyline schema


labkey-jeckels:
Let's hold on merging until this test suite has completed:

https://teamcity.labkey.org/viewQueued.html?itemId=773213

vagisha:
I was using double instead of Double in MeasuredDriftTime for the new nullable fields.  I will fix that and create another PR.  Closing this one without merging.

PR #21, opened 2019-06-30
Fix TargetedMSQCSummaryTest.testShowAutoQC and TargetedMSQCTest.testQ…

labkey-jeckels:
…CPlotExclusions

Update test based on new QC summary layout (missed a couple of test cases with original refactor) and there's a default guide set now to generate outliers even if user hasn't explicitly created one


labkey-jeckels:
I had missed that these failures were assigned to me. Trying to get them cleaned up before we branch for 19.2

PR #25, opened 2019-07-09
Issue 37921: Use exp.data to check for completeness of raw data instead of checking for existence of files

vagisha:
Issue 37921: Use exp.data to check for completeness of raw data instead of checking for existence of files

Lookup exp.data to determine if all the raw data has been uploaded. 
Data for an experiment can be organized in sub-containers. This patch eliminates checking for the existence of raw files in the parent container if they are not found in the same container as the Skyline document. Raw data for Skyline documents in a sub-container has be be uploaded to the same container.


vagisha:
Thanks for the review!  I added unit tests: https://github.com/LabKey/targetedms/commit/b1a2754250773dca1c68d7091ac5bb64729c9468

PR #26, opened 2019-07-09
Issue 37919: Include only those modifications in ProteomeXchange XML that are actually used in the Skyline document  

vagisha:
<peptide_modification> under <peptide_settings> in a .sky file includes all the modifications that the user checked in the Peptide Settings > Modifications tab. This can include more modifications than are actually used in the document, i.e. there is a peptide in the document with that modification. When validating a ProteomeXchange submission only include those modifications that were found in at least one peptide in the document.

### DIFF
--- a/src/org/labkey/targetedms/query/ModificationManager.java
+++ b/src/org/labkey/targetedms/query/ModificationManager.java
@@ -23,6 +23,7 @@ import org.labkey.api.data.SQLFragment;
 import org.labkey.api.data.Selector;
 import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.data.SqlSelector;
+import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.TableSelector;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.util.Pair;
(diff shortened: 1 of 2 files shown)


src/org/labkey/targetedms/query/ModificationManager.java
```diff
@@ -156,6 +157,40 @@ public void exec(ResultSet rs) throws SQLException
         return new SqlSelector(TargetedMSManager.getSchema(), sql).getArrayList(PeptideSettings.RunStructuralModification.class);
     }
 
+    public static List<PeptideSettings.IsotopeModification> getIsotopeModificationsUsedInRun(int runId)
```
labkey-jeckels: Can you add comments to the "UsedInRun" and "ForRun" variants of the methods to help distinguish them? It's mostly apparent from the method names, but since it's such a subtle difference would be good to be a bit more explicit.

vagisha: Comments added in commit https://github.com/LabKey/targetedms/commit/3ea4aa9a058c4bf38cbd107c3af7cf85bc30e16e

PR #29, opened 2019-07-10
Audit log viewer

rita-gwen:
This is implementation audit log deletion logic as part of a document version deletion and first version of the audit log viewer page.
There are integration tests for the deletion but no UI tests for the viewer yet.
Unfortunately there is another schema change. I assumed that this code won't make it to the 19.1 release so I set the module version to 19.20. Let me know if this should be changed.


resources/schemas/dbscripts/postgresql/targetedms-19.193-19.20.sql
```diff
@@ -0,0 +1,3 @@
+
+
+ALTER TABLE targetedms.Runs ADD COLUMN AuditLogEntriesCount INT4 DEFAULT 0 NOT NULL;
```
labkey-jeckels: This will be the first script of 19.20. I believe develop was just recently bumped to set module versions to 19.20. Please make it 19.20-19.21.

src/org/labkey/targetedms/SkylineAuditLogManager.java
```diff
@@ -108,30 +134,38 @@ public boolean verifyPreRequisites() throws AuditLogException{
...
-        BaseSelector.ResultSetHandler<Integer> resultSetHandler = (rs, conn) -> {
-            if(rs.next()){
-                return rs.getInt("docCount");
-            }
-            else
-                return 0;
-        };
-        Integer docCount = new SqlExecutor(runsTbl.getSchema()).executeWithResults(query, resultSetHandler);
-        _logTree = buildLogTree();
+        SimpleFilter.AndClause and = new SimpleFilter.AndClause();
```
labkey-jeckels: This should work fine, but I don't think you need the AndClause. You can just append all of these directly to docFilter, since all of the clauses get AND'd together.

AndClause is useful when you're composing a more complex set of criteria into a single clause, often in conjunction with OrClause.

src/org/labkey/targetedms/TargetedMSModule.java
```diff
@@ -207,7 +207,7 @@ public String getName()
     @Override
     public double getVersion()
     {
-        return 19.193;
+        return 19.20;
```
labkey-jeckels: Set to 19.21, per comment on the SQL script above.

src/org/labkey/targetedms/parser/skyaudit/AuditLogEntry.java
```diff
@@ -246,6 +251,36 @@ public AuditLogEntry persist(){
...
+     * @return this instance if the update is successful.
+     */
+    public AuditLogEntry updateVersionId(Integer pVersionId){
+        this.setVersionId(pVersionId);
+        SQLFragment sqlUpdate = new SQLFragment("");
+
+        if(this._entryId == null){  //entryId can be null if this is an existing entry read from a file and not persisted into the database
+            //attempting to retrieve entryId using the alternative key of document GUID and entry hash.
+            sqlUpdate.append("UPDATE targetedms.AuditLogEntry SET versionId = ? WHERE documentGUID = ? AND entryHash = ?");
+            sqlUpdate.addAll(List.of(pVersionId, this._documentGUID.toString(), this._entryHash));
```
labkey-jeckels: Don't think the List.of() should be required - can just call the vararg variant of addAll() directly

src/org/labkey/targetedms/parser/skyaudit/DatabaseUtil.java
```diff
@@ -0,0 +1,50 @@
...
+import org.labkey.api.data.SqlExecutor;
+import org.labkey.targetedms.TargetedMSSchema;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class DatabaseUtil
+{
+
+    public static Object retrieveSimpleType(SQLFragment pQuery){
```
labkey-jeckels: Consider using SqlSelector instead of SqlExecutor, where you should be able to call new SqlSelector().getObject(Integer.class) and SqlSelector.getMap()

rita-gwen:
I had to rebase my local branch to make sure 19.20-19-21 script is applied correctly during the server startup (because all 19.20 scripts got consolidated into a single file). Now I cannot push the fixes into the same remote branch (audit_log_viewer) because local and remote have different fork point. There are three options:
 - ask you to rebase it to 14ab23b9.
 - rebase it myself with --force option
 - create a different branch (and different pull request as a result)

Which option would you prefer?

labkey-jeckels:
@labkey-tchad Trey, you're much savvier with Git merging than me. Can you advise? See Rita's comments above.

And Rita, for future reference, please use our Git feature branch naming conventions so TeamCity can provide build and test coverage:

https://www.labkey.org/Documentation/wiki-page.view?name=git_branch_naming

rita-gwen:
OK, I pushed this commit from the original local branch but I cannot test it there properly. After the merge into develop it should work correctly. Sorry, this schema numbering method is awkward.

rita-gwen:
Sorry I did not see the branch naming convention before, I'll follow it from now on. I usually don't have issue or scrumwise ID, but the rest is fine.

labkey-jeckels:
Yes, the IDs aren't important from TeamCity's perspective, so they're fine to omit for your work.

labkey-tchad:
I added a new branch to verify this on TeamCity (fb_audit_log_viewer).
Tests are queued:
https://teamcity.labkey.org/viewQueued.html?itemId=780451
https://teamcity.labkey.org/viewQueued.html?itemId=780453

labkey-tchad:
Having problems during deletion:
```
java.lang.AssertionError
       at org.labkey.targetedms.SkylineAuditLogManager.deleteDocumentVersionLog(SkylineAuditLogManager.java:355)
       at org.labkey.targetedms.TargetedMSManager.deleteRuns(TargetedMSManager.java:1099)
       at org.labkey.targetedms.TargetedMSManager.deleteRun(TargetedMSManager.java:724)
       at org.labkey.targetedms.TargetedMSDataHandler.deleteData(TargetedMSDataHandler.java:125)
       at org.labkey.experiment.api.ExperimentServiceImpl.deleteExperimentRunsByRowIds(ExperimentServiceImpl.java:3424)
       at org.labkey.experiment.controllers.exp.ExperimentController$DeleteSelectedExpRunsAction.deleteObjects(ExperimentController.java:2796)
       at org.labkey.experiment.controllers.exp.ExperimentController$AbstractDeleteAction.handlePost(ExperimentController.java:2855)
       at org.labkey.experiment.controllers.exp.ExperimentController$DeleteSelectedExpRunsAction.handlePost(ExperimentController.java:2748)
       at org.labkey.experiment.controllers.exp.ExperimentController$AbstractDeleteAction.handlePost(ExperimentController.java:2837)
```

rita-gwen:
Yes, I think I was too aggressive with that documentGUID != null assertion. There are still a lot of legacy documents with no GUID. I pushed the fix, but I will look at the problem closer and will try to reproduce it here.

rita-gwen:
Looks like this exception is gone now. There is a couple of tests still failing but they seem to be unrelated to audit log code.

labkey-jeckels:
@rita-gwen To answer your question about the server log, look under the Artifacts tab. There should be a .tar.gz with various Tomcat logs, including labkey.log. If a test fails, there will also be screenshots and HTML grabs of the browser immediately after the failure.

rita-gwen:
Yes, thank you, I found them already. 
Looks like there is another schema version numbering conflict. There are two branches trying to grab 19.21. Should I update mine to 19.22?

labkey-tchad:
@rita-gwen The failures now are pre-existing or intermittent, except for the `JUnitFooter` failure.
It attempts to delete the test container and gets this error:
```
Failed to copy property 'peptideGroupCount' on class org.labkey.targetedms.TargetedMSRun
```

rita-gwen:
I think this one is mine. I have an idea where it comes from. Should have a fix in half an hour.

rita-gwen:
OK, the fix is pushed, hopefully it should correct this problem.

rita-gwen:
Looks like the JUnitFooter problem is fixed. Let me know if anything else is required to merge this.

labkey-tchad:
@rita-gwen Please squash-merge at your earliest convenience. Be sure to rewrite the auto-generated commit message. It should read as if the entire feature was added in one commit (e.g. it doesn't need details about code review and other iteration)

PR #33, opened 2019-07-12
Fb run scoped metrics

labkey-ankurj:


webapp/TargetedMS/js/QCPlotHoverPanel.js
```diff
@@ -255,15 +256,19 @@ Ext4.define('LABKEY.targetedms.QCPlotHoverPanel', {
...
-                chromInfoId: this.pointData['PrecursorChromInfoId']
-            });
+        if(this.metricProps.precursorScoped) {
+            var action = this.pointData['dataType'] == 'Peptide' ? "precursorAllChromatogramsChart" : "moleculePrecursorAllChromatogramsChart",
+                    url = LABKEY.ActionURL.buildURL('targetedms', action, LABKEY.ActionURL.getContainer(), {
+                        id: this.pointData['PrecursorId'],
+                        chromInfoId: this.pointData['PrecursorChromInfoId']
+                    });
+        } else {
+            url =  LABKEY.ActionURL.buildURL('targetedms', 'showPrecursorList', null)
```
labkey-jeckels: Does the run's ID get appended somewhere else?

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -858,7 +867,27 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
                         this.havePlotOptionsChanged = true;
+                        var items = this.otherPlotOptionsToolbar.items.items;
+                        var showAllSeriesCheckBoxLocation;
+
+                        for(var i=0; i<items.length;i++ ) {
+                            if(items[i].boxLabel ==='Show All Series in a Single Plot')
+                                showAllSeriesCheckBoxLocation = i;
+                        }
 
+                        if(this.getMetricPropsById(newVal).precursorScoped) {
```
labkey-jeckels: Using setVisible() would be a little less code than needing a separate sets of hide() and show() calls.

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -417,6 +417,15 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
                 padding: '0 10px 10px 10px',
                 items: toolbarItems
             });
+
+            //hiding the show All series in a single plot checkbox for run scoped metrics
+            var items = this.otherPlotOptionsToolbar.items.items;
```
labkey-jeckels: Do we need to reshow later? Or does this get completely reinitialized so there's no need?

src/org/labkey/targetedms/TargetedMSModule.java
```diff
@@ -207,7 +207,7 @@ public String getName()
     @Override
     public double getVersion()
     {
-        return 19.193;
+        return 19.21;
```
labkey-jeckels: There is another PR that's going to be merged before this one that also has an upgrade script. Please update this PR's version to 19.22.

labkey-jeckels: https://github.com/LabKey/targetedms/pull/29#

webapp/TargetedMS/js/QCTrendPlotPanel.js
labkey-ankurj: Yes, we need to reshow this when molecule scoped metric is selected from the dropdown. It does not get reinitialized.

labkey-jeckels: Are all of the shows and hides operating on the same components? It seemed odd to me that we have two sets of hide() calls and only one set of show(). And that they span indices -1 to +2, and -2 to +1.

labkey-ankurj: Yes, they are operating on same components and the range includes the separators and tab spacers associated with the checkbox.

src/org/labkey/targetedms/TargetedMSModule.java
labkey-ankurj: done

labkey-ankurj:
Adding Chris as a reviewer for automated test.

test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
```diff
@@ -237,7 +237,8 @@ public String getCurrentEndDate()
         LHRATIO("Light/Heavy Ratio", false),
         TPAREARATIO("Transition/Precursor Area Ratio", true),
         TPAREAS("Transition & Precursor Areas", true),
-        MASSACCURACTY("Mass Accuracy", true);
+        MASSACCURACTY("Mass Accuracy", true),
```
labkey-chrisj: eventually, we should change MASSACCURACTY to be MASSACCURACY (lol one of these has a typo in it)

PR #35, opened 2019-07-17
Adding missing column for targetedms.runs in xml

labkey-ankurj:


rita-gwen:
Sorry I forgot about it. Would it be possible to run this test as a part of the local build process to detect the problem earlier?

labkey-jeckels:
@rita-gwen Yes. On your local server, you can run the integration tests via the web UI as described here:

https://www.labkey.org/Documentation/wiki-page.view?name=runningTests

The one that checks for discrepancies between the XML metadata and the DB schema is org.labkey.core.admin.test.SchemaXMLTestCase

PR #38, opened 2019-07-26
Fb custom metric

labkey-ankurj:


src/org/labkey/targetedms/query/QCMetricConfigurationTable.java
```diff
@@ -42,6 +49,19 @@ protected void applyContainerFilter(ContainerFilter filter)
         super.applyContainerFilter(filter);
     }
 
+    @Override
+    public boolean hasPermission(@NotNull UserPrincipal user, @NotNull Class<? extends Permission> perm)
+    {
+        Class<? extends Permission> permissionToCheck = perm == ReadPermission.class ? ReadPermission.class : AdminPermission.class;
+        return getContainer().hasPermission(user, permissionToCheck);
```
labkey-ankurj: @labkey-jeckels Am I doing the correct permission check here or should it be just - return getContainer().hasPermission(user, perm); ?

labkey-jeckels: That looks good, as we want to make it so that only admins can curate the set of metrics for a given container.

PR #43, opened 2019-08-15
Prevent calculating fold changes on decoy peptides.

nickshulman:
There is a Skyline document that takes a really long time to import into Panorama.
This was happening because Panorama is trying to calculate protein-level fold changes, and there is one "protein" that has thousands of decoy peptides in it.
Part of the fold change calculation is order N-squared in the total number of transitions under the protein, so this was taking a really long time.

This change makes it so that we do not include decoy peptides when doing protein-level fold change calculations. This will fix the symptom we are seeing where this document took a long time to import.

Longer term, I should also fix it so that we don't have that order-n-squared problem, so that it won't be slow even if a real protein has lots of transitions.

Vagisha has suggested that we merge this into release19.2-SNAPSHOT?


vagisha:
Hi Nick,

Here is a link to LabKey's documentation on branch naming:
https://www.labkey.org/Documentation/wiki-page.view?name=git_branch_naming.
I think this ensures that the branch gets built and tested on TeamCity.

Thanks,
Vagisha

labkey-jeckels:
Already fixed with this PR: https://github.com/LabKey/targetedms/pull/45

PR #44, opened 2019-08-15
Issue 38270: Incorrect x-axis location of QC annotation

vagisha:


labkey-jeckels:
Both test failures look to be intermittent issues that have happened on develop as well, so cleared to merge.

PR #65, opened 2019-09-19
Pipeline job-based crawler to log the status of all Panorama chromatograms and their availability in the DB and from the file system

labkey-jeckels:


src/org/labkey/targetedms/parser/PrecursorChromInfo.java
```diff
@@ -450,26 +458,54 @@ public Chromatogram createChromatogram(TargetedMSRun run)
...
-                            compressedBytes = onDemandBytes;
+                            LOG.debug("Attempting to fetch chromatogram bytes (possibly cached) from " + skydPath + " for PrecursorChromInfo " + _generalMoleculeChromInfoId);
+                            byte[] diskBytes = ON_DEMAND_CHROM_CACHE.get(new Tuple3<>(skydPath, _chromatogramOffset, _chromatogramLength));
+                            if (diskBytes == null)
+                            {
+                                status = Chromatogram.SourceStatus.skydMissing;
+                            }
+                            else if (!Arrays.equals(databaseBytes, diskBytes))
+                            {
+                                LOG.error("Chromatogram bytes for PrecursorChromInfo " + _generalMoleculeChromInfoId + " do not match between .skyd and DB. Using database copy. Lengths: " + (diskBytes == null ? "null" : Integer.toString(diskBytes.length)) + " vs " + databaseBytes.length);
```
labkey-ankurj: diskBytes == null is always null here.

labkey-jeckels: True, and databaseBytes could have been null. Update pushed.

labkey-ankurj: Looks good, thanks.

PR #69, opened 2019-09-23
Cleanup import codepath

labkey-jeckels:
Don't creating an inner transaction to help improve error reporting, fixing a problem Vagisha was seeing.

Pretty big refactor of the previously massive (400+ line) importSkylineDoc() method.

Tests are happy!

### DIFF
--- a/src/org/labkey/targetedms/SkylineDocImporter.java
+++ b/src/org/labkey/targetedms/SkylineDocImporter.java
@@ -261,10 +261,10 @@ public class SkylineDocImporter
             run.setiRTscaleId(insertiRTData(parser));
 
             // 1. Transition settings
-            OptimizationSettings optimizationSettings = insertTransitionSettings(parser.getTransitionSettings());
+            OptimizationInfo optimizationInfo = insertTransitionSettings(parser.getTransitionSettings());
 
             // 2. Replicates and sample files
-            ReplicateInfo replicateInfo = insertReplicates(run, parser, optimizationSettings, folderType);
+            ReplicateInfo replicateInfo = insertReplicates(run, parser, optimizationInfo, folderType);
 
             // 3. Peptide settings
             PeptideSettings pepSettings = parser.getPeptideSettings();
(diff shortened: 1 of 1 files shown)


src/org/labkey/targetedms/SkylineDocImporter.java
```diff
@@ -233,465 +234,472 @@ private void importSkylineDoc(TargetedMSRun run, File f) throws XMLStreamExcepti
...
+            }
+            if(optimizationSettings._dpPredictor != null && skyReplicate.getDpPredictor() != null && optimizationSettings._dpPredictor.equals(skyReplicate.getDpPredictor()))
+            {
+                replicate.setDpPredictorId(optimizationSettings._dpPredictor.getId());
+            }
 
-                    if (folderType == TargetedMSService.FolderType.QC)
+            if (folderType == TargetedMSService.FolderType.QC)
+            {
+                // In QC folders insert a replicate only if at least one of the associated sample files will be inserted
```
vagisha: This comment should be removed, I think.  We now keep replicates and sample files from the latest uploaded Skyline document and delete older ones.  This comment (and the one on line 1939) look like they are from before this change.

src/org/labkey/targetedms/SkylineDocImporter.java
```diff
@@ -233,465 +234,472 @@ private void importSkylineDoc(TargetedMSRun run, File f) throws XMLStreamExcepti
...
 
-                // Store the document settings
-                // 0. iRT information
-                run.setiRTscaleId(insertiRTData(parser));
+            // Store the document settings
+            // 0. iRT information
+            run.setiRTscaleId(insertiRTData(parser));
 
+            // 1. Transition settings
+            OptimizationSettings optimizationSettings = insertTransitionSettings(parser.getTransitionSettings());
```
vagisha: Perhaps rename OptimizationSettings to OptimizationInfo to be consistent with ReplicateInfo, ModificationInfo etc.?

brendanx67:
I hope that the tests are also comprehensive.

src/org/labkey/targetedms/SkylineDocImporter.java
labkey-jeckels: I think those comments are still relevant. While we retain older data, we still need to insert newly added replicates and sample files.

PR #70, opened 2019-09-24
Support for new skyd file format with SampleId

nickshulman:
We just committed a change to Skyline which bumps the version number of the .skyd file to 14.
This change is not backwards compatible, so we need to update Panorama to be able to handle the new format.

We probably also want to add these columns to the database schema. Maybe someone else will want to do that work. You can read the SampleId and Instrument Serial Number from the .sky file. The attributes on the "<sample_file>" element are "sample_id" and "instrument_serial_number"


nickshulman:
The SampleId is something that the user can type into the Thermo software
when they are setting up their method.

Ideally we wouldn't store these values in the .skyd file at all-- we would
just have them in the .sky file. However, because of the way that
chromatogram extraction works in Skyline, we end up having to write them to
the .skyd file, and then they get copied into the .sky file later. It's the
same way as we were already handling things like "Acquired Time".
-- Nick

labkey-jeckels:
OK. It sounds like we should wait on the Panorama side until this is getting written into the .sky XML. Let me know if I've misunderstood. I've just merged this update in the meantime.

nickshulman:
It does get written to the .sky file. If you look at the .sky file in the
"SampleIdTest.sky.zip" zip file, you'll see the XML attributes:

It looks like:
        <sample_file id="S_1_f0"
file_path="E:\skydata\20110215_MikeB\S_1.RAW?centroid_ms1=true"
sample_name="S_1" *sample_id="1:A,1" instrument_serial_number="6147F"*
acquired_time="2010-11-24T06:37:50"
modified_time="2011-02-22T20:23:31" tic_area="2531654400"
ion_mobility_type="compensation_V">
-- Nick

labkey-jeckels:
Oh, cool. Thanks.

https://github.com/LabKey/targetedms/pull/74

PR #73, opened 2019-09-25
Issue 38591: Use attributegroupingid value if present for PeptideModi…

labkey-jeckels:
…fiedAreaProportion and PrecursorModifiedAreaProportion calculations


test/src/org/labkey/test/tests/targetedms/TargetedMSExperimentTest.java
```diff
@@ -54,7 +65,45 @@ public void testSteps()
 
         //small molecule
         importData(SKY_FILE_SMALLMOL_PEP, 3);
-        verifyImportedSmallMoleculeData();
+//        verifyImportedSmallMoleculeData();
```
labkey-ankurj: remove?

labkey-jeckels: I just re-enabled this. I had removed it to make it faster to test the updates.

labkey-jeckels:
Tests running now:

https://teamcity.labkey.org/viewLog.html?buildId=823466&buildTypeId=LabkeyTrunk_ModuleSuites_TargettedMSPostgres

PR #77, opened 2019-10-03
Fb skyline lists

nickshulman:
Here is a first crack at support for Skyline lists.

I have a test file "ListTest.sky.zip" that has some examples of lists in it with really long column names, etc. It is possible to use the Query module to write queries against lists with really long column names, but you might run into problems with GET URLs being too long.

There is a lot more list functionality that we should try to get supported. Specifically, annotations can now have a "lookup" attribute which indicates that they should join in columns from a List.
(This pull request just had the minimal amount of work to get this data into the database).


src/org/labkey/targetedms/parser/list/ListItemValue.java
```diff
@@ -0,0 +1,92 @@
...
+        _numericValue = numericValue;
+    }
+
+    public void setValue(Object value)
+    {
+        if (value == null)
+        {
+            setTextValue(null);
+            setNumericValue(null);
+            return;
```
labkey-jeckels: Seems like this would be cleaner as if/else instead of bunch of if/returns

resources/schemas/dbscripts/postgresql/targetedms-19.25-19.26.sql
```diff
@@ -0,0 +1,58 @@
...
+    CONSTRAINT PK_List PRIMARY KEY(Id),
+    CONSTRAINT FK_List_RunId FOREIGN KEY(RunId) REFERENCES targetedms.Runs(Id)
+);
+CREATE TABLE targetedms.ListColumnDefinition
+(
+    Id SERIAL NOT NULL,
+    ListDefinitionId INT NOT NULL,
+    ColumnIndex INT NOT NULL,
+    AnnotationType VARCHAR(20) NOT NULL,
+    Name TEXT NOT NULL,
```
labkey-jeckels: Does this really need to be unlimited? I see the code that is conditional on SQL Server for this. Does anyone need list columns that are more than a few hundred characters? Do these get autogenerated?

src/org/labkey/targetedms/query/SkylineListManager.java
```diff
@@ -0,0 +1,126 @@
...
+{
+    /**
+     * Returns true if the long text columns such as ListDefinition.Name support the equals operator.
+     */
+    public static boolean canFilterNameColumn()
+    {
+        return !TargetedMSSchema.getSchema().getSqlDialect().isSqlServer();
+    }
+
+    public static ListDefinition saveListData(User user, ListData listData) {
```
labkey-jeckels: Please reformat with brace on the next line

src/org/labkey/targetedms/query/SkylineListManager.java
```diff
@@ -0,0 +1,126 @@
...
+        return TargetedMSSchema.getSchema().getTable(TargetedMSSchema.TABLE_LIST_COLUMN_DEFINITION);
+    }
+
+    public static List<ListDefinition> getListDefinitions(Container container) {
+        SQLFragment fragment = new SQLFragment("SELECT * FROM ");
+        fragment.append(getListDefTable(), "t");
+        fragment.append(" WHERE t.RunId IN (SELECT Id FROM ");
+        fragment.append(TargetedMSSchema.getSchema().getTable(TargetedMSSchema.TABLE_RUNS), "r");
+        fragment.append(new SQLFragment(" WHERE Container = ? AND StatusId = ? AND deleted = ?)",
+                container.getId(), 1, Boolean.FALSE));
```
labkey-jeckels: Please use constant from SkylineDocImporter instead of hard-coding 1

src/org/labkey/targetedms/query/SkylineListTable.java
```diff
@@ -0,0 +1,88 @@
...
+    {
+        SQLFragment fragment = new SQLFragment("SELECT Id FROM targetedms.ListItem WHERE ListDefinitionId = ?", _listDefinition.getId());
+        return fragment;
+    }
+
+    class ListColumnInfo extends BaseColumnInfo
+    {
+        ListColumn _listColumn;
+        public ListColumnInfo(SkylineListTable listTable, ListColumn listColumn) {
+            super(new FieldKey(null, listColumn.getName()), listTable);
```
labkey-jeckels: Seems like this should set the column type somewhere so that we know it's a string, boolean, or number, right?

labkey-jeckels:
Just merged in changes from the develop branch to make the test compilation happy. If TeamCity is satisfied (tests are running now) we can merge this PR tomorrow.

PR #84, opened 2019-10-21
Issue 38721: metric label encoding for QC plot select and QC summary tooltip

cnathe:


cnathe:
TeamCity run of MS2 and TargetedMS suites all passed.

PR #87, opened 2019-10-21
Fix the version support number returned by GetMaxSupportedVersionsAction

nickshulman:
Here is the fix to make it so that Skyline queries the server about which skyd file format to save as.

There is a different problem in Skyline where trying to save as version 19.1 produces something invalid. This will be fixed in Skyline github project in pull request number 685.


labkey-jeckels:
We just cut version 19.2.7 with the hotfix to support SKYD 14. Unless we discover another critical problem, our next patch release for 19.2.x will only be a week or two ahead of the 19.3.0 release, so I vote for keeping this targeted at develop

PR #90, opened 2019-10-23
Remove Panorama Public code from targetedms

vagisha:


vagisha:
Added constants to service interface instead of methods. 
Attempting to merge to a feature branch so tests can run.

PR #92, opened 2019-10-23
Issue 38721: SQL parser error in the TargetedMSController.GetQCMetricOutliersAction

labkey-ankurj:


src/org/labkey/targetedms/outliers/CUSUMOutliers.java
```diff
@@ -111,42 +118,46 @@ private String getEachSeriesTypePlotDataSql(String type, int id, String schemaNa
...
                 String schema2Name = configuration.getSeries2SchemaName();
                 String query2Name = configuration.getSeries2QueryName();
                 label = configuration.getSeries2Label();
-                sqlBuilder.append(sep).append("(").append(getEachSeriesTypePlotDataSql("series2", id, schema2Name, query2Name, where, label)).append(")");
+                sqlBuilder.append(sep).append("(").append(getEachSeriesTypePlotDataSql("series2", id, schema2Name, query2Name, where, METRIC_LABEL_TWO)).append(")");
             }
         }
-        return "SELECT * FROM (" + sqlBuilder.toString() + ") a"; //wrap unioned results in sql to support sorting
+        sqlBuilder.append(") a");
+        return  sqlBuilder.toDebugString() ; //wrap unioned results in sql to support sorting
```
labkey-jeckels: This method is for debugging purposes - probably want to use getSQL() or just return SQLFragment directly

src/org/labkey/targetedms/outliers/Outliers.java
```diff
@@ -39,16 +50,78 @@
...
     }
+
+    public static Map<String, Object> getNamedParametersForQCConfigurations()
+    {
+        return _namedParameters;
+    }
+
+    public static void setNamedParametersForQCConfigurations(List<QCMetricConfiguration> configurations)
+    {
+        Map<String, Object> namedParameters = new HashMap<>();
```
labkey-jeckels: This shouldn't be static and stored as a variable - what if two different users are querying different folders at the same time?

src/org/labkey/targetedms/outliers/Outliers.java
```diff
@@ -39,16 +50,78 @@
...
+        configurations.forEach(config -> {
+            String nameKey = METRIC_NAME + config.getId() + "$";
+            namedParameters.put(nameKey, config.getName());
+
+            if (null != config.getSeries1Label())
+            {
+                String labelKey = METRIC_LABEL_ONE + config.getId() + "$";
```
labkey-jeckels: What if we make this part of the query so that it fetches the name from the config table itself? That way the SQL would only need to include the RowId.

PR #95, opened 2019-10-29
Issue 38800: Fix parsing of attributes in SkylineDocumentParser.readGeneralTransition()

labkey-jeckels:


labkey-jeckels:
I swear I based this on develop, but it looks like it's not.

PR #96, opened 2019-10-29
Issue 38800: Fix parsing of attributes in SkylineDocumentParser.readGeneralTransition()

labkey-jeckels:
Take 2, hopefully with the right diff now.

### DIFF
--- a/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
+++ b/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
@@ -2221,12 +2221,6 @@ public class SkylineDocumentParser implements AutoCloseable
         if(explicitDeclusteringPotential != null)
             transition.setExplicitDeclusteringPotential(Double.valueOf(explicitDeclusteringPotential));
 
-        String explicitDriftTimeMsec = reader.getAttributeValue(null, "explicit_ion_mobility_high_energy_offset");
-        if (explicitDriftTimeMsec == null)
-            explicitDriftTimeMsec = reader.getAttributeValue(null, "explicit_drift_time_msec");
-        if(explicitDriftTimeMsec != null)
-            transition.setExplicitDriftTimeMSec(Double.valueOf(explicitDriftTimeMsec));
-
         String explicitDriftTimeHighEnergyOffsetMsec = reader.getAttributeValue(null, "explicit_ion_mobility_high_energy_offset");
         if (explicitDriftTimeHighEnergyOffsetMsec == null)
             explicitDriftTimeHighEnergyOffsetMsec = reader.getAttributeValue(null, "explicit_drift_time_high_energy_offset_msec");
--- a/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
+++ b/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
@@ -2200,11 +2200,16 @@ public class SkylineDocumentParser implements AutoCloseable
         if(explicitCollisionEnergy != null)
             transition.setExplicitCollisionEnergy(Double.valueOf(explicitCollisionEnergy));
 
-        String sLens = reader.getAttributeValue(null, "s_lens");
+        String sLens = reader.getAttributeValue(null, "explicit_s_lens");
+        if (sLens == null)
+            // Fall back on older attribute name
+            sLens = reader.getAttributeValue(null, "s_lens");
         if(sLens != null)
             transition.setsLens(Double.valueOf(sLens));
 
-        String coneVoltage = reader.getAttributeValue(null, "cone_voltage");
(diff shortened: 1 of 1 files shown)


src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -2202,27 +2202,27 @@ private void readGeneralTransition(XMLStreamReader reader, GeneralTransition tra
 
         String sLens = reader.getAttributeValue(null, "s_lens");
```
vagisha: Should be changed to "explicit_s_lens".  This is the new name of the attribute.  It used to be "s_lens" when this attribute was still associated with general_transition_group_type (precursor).  This info is in the Word doc.

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -2202,27 +2202,27 @@ private void readGeneralTransition(XMLStreamReader reader, GeneralTransition tra
...
         if(sLens != null)
-            transition.setExplicitCollisionEnergy(Double.valueOf(sLens));
+            transition.setsLens(Double.valueOf(sLens));
 
         String coneVoltage = reader.getAttributeValue(null, "cone_voltage");
```
vagisha: Should be changed to "explicit_cone_voltage".  Same reason as above.

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
labkey-jeckels: Thanks. I didn't read far enough!

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -2202,27 +2202,27 @@ private void readGeneralTransition(XMLStreamReader reader, GeneralTransition tra
...
         if(explicitDeclusteringPotential != null)
-            transition.setExplicitCollisionEnergy(Double.valueOf(explicitDeclusteringPotential));
+            transition.setExplicitDeclusteringPotential(Double.valueOf(explicitDeclusteringPotential));
 
         String explicitDriftTimeMsec = reader.getAttributeValue(null, "explicit_drift_time_msec");
         if(explicitDriftTimeMsec != null)
-            transition.setExplicitCollisionEnergy(Double.valueOf(explicitDriftTimeMsec));
+            transition.setExplicitDriftTimeMSec(Double.valueOf(explicitDriftTimeMsec));
 
         String explicitDriftTimeHighEnergyOffsetMsec = reader.getAttributeValue(null, "explicit_drift_time_high_energy_offset_msec");
```
vagisha: Should read attribute "explicit_ion_mobility_high_energy_offset".  Same reason as above.  It used to be called "explicit_drift_time_high_energy_offset_msec" when it was in general_transition_group_type (precursor).

labkey-jeckels:
@vagisha Or should I remove the old attribute names completely, because they will never be present in this part of the XML?

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -2200,29 +2200,38 @@ private void readGeneralTransition(XMLStreamReader reader, GeneralTransition tra
...
 
         String explicitDeclusteringPotential = reader.getAttributeValue(null, "explicit_declustering_potential");
         if(explicitDeclusteringPotential != null)
-            transition.setExplicitCollisionEnergy(Double.valueOf(explicitDeclusteringPotential));
+            transition.setExplicitDeclusteringPotential(Double.valueOf(explicitDeclusteringPotential));
 
-        String explicitDriftTimeMsec = reader.getAttributeValue(null, "explicit_drift_time_msec");
+        String explicitDriftTimeMsec = reader.getAttributeValue(null, "explicit_ion_mobility_high_energy_offset");
+        if (explicitDriftTimeMsec == null)
+            explicitDriftTimeMsec = reader.getAttributeValue(null, "explicit_drift_time_msec");
```
vagisha: "explicit_drift_time_msec" is not at the transition level.  It used to be at the precursor level and was renamed to "explicit_ion_mobility" there.  So it does not make sense reading this for a transition.  explicitDriftTimeMsec column can be deleted from GeneralTransition in 20.1.

labkey-jeckels: OK, deleted with latest push.

vagisha:
> @vagisha Or should I remove the old attribute names completely, because they will never be present in this part of the XML?

Yes, I don't expect the old attribute names to be present at the transition level in the current XML.  So they can be removed. There will be older documents where the old names will show up at the precursor level. I don't know if you want to read them at the precursor level but save them in the transitions table!

PR #111, opened 2019-12-06
Issue 39149: NPE when trying to view chromatograms

vagisha:
NPE is thrown when trying to view chromatograms for a Skyline document imported into a folder after setting the module property "Skip chromatogram import into database" to "Enabled". On PanoramaWeb this property is enabled site-wide.


vagisha:
Closing PR since this will have to be merged with the release19.3 branch instead

PR #115, opened 2019-12-11
Get chromatogram data from skyd file if not found in db when writing chromatogram library

vagisha:
If chromatograms for a precursor are not saved in the database, due to "Skip chromatogram import into database" property being enabled, values in the "Chromatogram" column of the "Precursor" table in a chromatogram library are null. Chromatogram data should be read from the skyd file in this case.

### DIFF
--- a/src/org/labkey/targetedms/parser/PrecursorChromInfo.java
+++ b/src/org/labkey/targetedms/parser/PrecursorChromInfo.java
@@ -445,59 +445,9 @@ public class PrecursorChromInfo extends ChromInfo<PrecursorChromInfoAnnotation>
 
             ChromatogramBinaryFormat binaryFormat = ChromatogramBinaryFormat.values()[getChromatogramFormat()];
 
-            byte[] databaseBytes = getChromatogram();
-            byte[] compressedBytes = databaseBytes;
-
-            Chromatogram.SourceStatus status;
-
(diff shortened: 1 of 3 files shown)


src/org/labkey/targetedms/parser/PrecursorChromInfo.java
```diff
@@ -445,71 +445,86 @@ public Chromatogram createChromatogram(TargetedMSRun run, boolean loadFromSkyd)
...
+            byte[] uncompressedBytes = SkylineBinaryParser.uncompressStoredBytes(compressedBytes, getUncompressedSize(), _numPoints, _numTransitions);
+            return binaryFormat.readChromatogram(uncompressedBytes, _numPoints, _numTransitions, status);
+        }
+        catch (IOException | DataFormatException exception)
+        {
+            throw new UnexpectedException(exception);
+        }
+    }
+
+    private Object[] getCompressedBytesAndStatus(TargetedMSRun run, boolean loadFromSkyd)
```
labkey-jeckels: Let's make a class to hold the return value here, or at least use a Pair<byte[], Chromatogram.SourceStatus> instead of an untyped Object array.

vagisha: I added a class to hold the return values.

PR #118, opened 2019-12-13
Add page titles

labkey-bpatel:


src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -4122,7 +4128,7 @@ else if(!useMin && (conflictRunId == cProtein.getNewProteinRunId()))
         @Override
         public NavTree appendNavTrail(NavTree root)
         {
-            return root;
+            return (null != getPageConfig().getTitle() ? root.addChild(getPageConfig().getTitle()) : root.addChild("Protein Conflicts"));
         }
```
labkey-klum: Probably doesn't hurt anything but unless the page title is set directly in getView or this is a base class for other actions, the page title should be null when appendNavTrail is called

PR #129, opened 2020-01-07
Correct values and tighten tolerances to fix TargetedMSQCGuideSetTest

labkey-jeckels:
This test fails locally for me on one of the drag/drop operations. I was too hasty to merge the attempted fix yesterday, which addressed the original failure but introduced a new, related one.

This PR should take care of that, and fix a new failure introduced by Cory's update of the tooltip text on bar plots.


cnathe (commented):
I'll defer to a tester on the review/approval, but the part about the fix for the bar plot hover text display looks good (@labkey-jeckels thanks for fixing this use case up for me).

labkey-danield (approved):

PR #130, opened 2020-01-14
Issue 39401: NPE while importing a QC document that has more than one replicate for the same raw file

vagisha:
Waiting for tests to run: https://teamcity.labkey.org/viewQueued.html?itemId=888278&tab=queuedBuildOverviewTab

### DIFF
--- a/src/org/labkey/targetedms/SkylineDocImporter.java
+++ b/src/org/labkey/targetedms/SkylineDocImporter.java
@@ -476,14 +476,14 @@ public class SkylineDocImporter
 
     private void deleteOldSampleFiles(ReplicateInfo replicateInfo)
     {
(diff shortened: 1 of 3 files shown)


labkey-jeckels (changes requested):
Can we implement equals() and hashCode() on SkylineEntity so that we can use a Set<SampleFile> instead of a Set<Integer>? Using Integer isn't bad, but SampleFile is a lot clearer in terms of what's being referenced.

I think this would work in SkylineEntity:

    @Override
    public boolean equals(Object o)
    {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        SkylineEntity that = (SkylineEntity) o;
        return _id == that._id;
    }

    @Override
    public int hashCode()
    {
        return Objects.hash(_id, getClass());
    }

labkey-jeckels (approved):

PR #139, opened 2020-02-08
Show full file size before clicking to see download options

labkey-jeckels:
Move to top level instead of just showing inside of the popup menu, per request from Brendan


brendanx67:
Can we get this into a patch that can be used to update PanoramaWeb.org?

labkey-jeckels:
Closing this one in favor of a 19.3 change.

https://github.com/LabKey/targetedms/pull/141

PR #145, opened 2020-03-01
Issue 38679: Duplicate rows get created for samples that have same samplefile name in multiple imported skyline documents for QC

labkey-ankurj:


src/org/labkey/targetedms/outliers/CUSUMOutliers.java
```diff
@@ -669,10 +669,11 @@ public JSONObject getSampleFilesJSON(Map<String, SampleFileInfo> sampleFiles)
 
         for (LJOutlier ljOutlier : ljOutliers)
         {
-            if(sampleFileInfo == null || (!(ljOutlier.getSampleFile() != null && ljOutlier.getSampleFile().equals(sampleFileInfo.getSampleFile()))))
+            String sampleFileString = ljOutlier.getSampleFile() + "_" + ljOutlier.getAcquiredTime();
```
labkey-jeckels, line 654: Why are you concatenating these together? To make sure it's unique? Would it be better to use the Id value from targetedms.samplefile?

src/org/labkey/targetedms/outliers/CUSUMOutliers.java
```diff
@@ -517,16 +520,17 @@ public int hashCode()
     }
 
 
-    private Map<String, Integer> processEachOutlier(boolean groupByGuideSet, Map<String, Integer> countObj, int guideSetId, Set<String> sampleFiles, String sampleFile)
+    private Map<String, Integer> processEachOutlier(boolean groupByGuideSet, Map<String, Integer> countObj, int guideSetId, Set<String> sampleFiles, String sampleFile, String sampleFileString)
```
labkey-jeckels: These arguments need some additional explanation. How do sampleFile and sampleFileString relate. Part of my confusion is that they're both of type String.

api-src/org/labkey/api/targetedms/model/SampleFileInfo.java
```diff
@@ -51,6 +52,16 @@ public void setSampleFile(String sampleFile)
         this.sampleFile = sampleFile;
     }
 
+    public int getSampleFileId()
```
labkey-jeckels: It looks like this got added but not used. See other comment about potentially using it instead of concatenating two other columns

labkey-jeckels (changes requested):

labkey-jeckels (changes requested):
Code looks good. Let's take this for 20.3. Can you update the PR to point to that branch?

labkey-ankurj:
I thought the tests will restart on 20.3-SNAPSHOT on teamcity after changing the base, but they didn't. Manually kicking them off before merging.

labkey-jeckels (approved):

PR #148, opened 2020-03-14
Issue 39961: Hide Targeted MS Chromatogram Crawler option when user doesn't have perm

labkey-jeckels:


labkey-adam (approved):
This seems okay, though if all we're displaying is a "you don't have permissions" message, then maybe just @Requires(ApplicationAdminPermission.class) on the action (and link registration) and call it good? @AdminConsoleAction is generally for cases where we want to show TroubleShooter something useful.

labkey-jeckels:
@labkey-adam 

I was going to do that originally, but saw this comment in ApplicationAdminPermission:

@Deprecated  // this will hopefully be swapped out for more fine-grained permissions soon; try to use those instead

I couldn't find any guidance on what else to use though.

labkey-jeckels:
Also, the link doesn't actually show for troubleshooters in the Admin Console because it's registered as needing AdminPermission. I verified that hitting the action gives the message to Troubleshooters, but allows App Admins.

labkey-adam:
Right, I saw that the link never appears for TroubleShooters, that's why it seemed odd to use @AdminConsoleAction. No one will see that "permissions" message, except if they're mucking with URLs; in which case, might as well use normal permission checking. But I don't really care that much...

labkey-jeckels:
I don't think it matter much for this use case either, but how should I pull this off without using ApplicationAdminPermission? Or should we undeprecate it?

labkey-adam:
I would just use it and undeprecate it. Fine-grained permissions are great for most cases, but many roles need a general perm like this.

labkey-jeckels:
Done. Probably didn't need to undeprecate in 20.3 but seems low risk

labkey-adam (approved):

PR #150, opened 2020-03-23
Skip importing TransitionChromInfos when they won't be useful

labkey-jeckels:
Specifically, skip for very large DIA documents (> 100,000 entries) and for data marked as a decoy, to reduce space usage in the relational DB that's unlikely to be useful


vagisha (approved):
Looks good.  Thanks.

labkey-ankurj (approved):

PR #166, opened 2020-04-17
Issue 40227: SQLException when importing a Skyline document with a chromatogram library

vagisha:


labkey-jeckels (approved):
I think TeamCity can get confused if the non-branch-specific part of the FB name overlaps with a FB it's also seen, which I'm guessing is why it's running against develop on one of the builds.

The other looks like it's cranking on 20.3 and should be done soon. Unless it turns up any problems, we should be able to merge this PR.

labkey-jeckels:
Looks like the Postgres side just kicked off against 20.3 too, so this PR no longer looks weird in terms of its test suites.

PR #167, opened 2020-04-22
Issue 40255: Skip serializing potentially huge, always unused QC data

labkey-jeckels:
… in TargetedMSController.GetQCMetricOutliersAction


labkey-ankurj (approved):
thank you for optimizing this, it is indeed faster from 9 sec to 4 sec for test data.

PR #173, opened 2020-05-05
Issue 40377: Support alternative to RowId value as URL parameter

labkey-jeckels:
#### Rationale
PanoramaPublic is a public repository for Skyline/Panorama data. Users upload data to a private folder, and there's a workflow to review, approve, and copy it to a public location.

https://panoramaweb.org/project/Panorama%20Public/begin.view?

Users sometimes create wikis that reference specific files they've imported. Those URLs current expect the Id column from the imported file, which is a RowId. That means that their link will break when it's copied to the public folder, since the data is reimported and will have a different RowId in its new home.

We can support an alternative URL parameter that lets the user specify the file name instead. If it's ambiguous, because the user uploaded the same file to different subdirectories under the file root (this should be exceedingly rare, as typically they're all in the file root itself), we'll simply treat it as if they didn't specify an ID at all.

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -3674,7 +3674,19 @@ public class TargetedMSController extends SpringActionController
     public static class RunDetailsForm extends QueryViewAction.QueryExportForm
     {
         private int _id = 0;
+        private String _fileName;
         private String _view;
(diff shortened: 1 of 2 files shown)


vagisha (approved):
Thanks!

labkey-susanh (approved):

PR #176, opened 2020-05-10
Improve QC summary outlier calculation perf

labkey-jeckels:
#### Rationale
As groups have loaded more and more data in Panorama QC folders, performance has become too slow. This eliminates two of the very expensive queries to calculate outliers in Java, and refactors the code to be simpler and avoid complex nested List/Map data structures.

It also simplifies the single major query that remains by doing joins outside of a union instead of for each subquery within the union

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1162


src/org/labkey/targetedms/model/GuideSetStats.java
```diff
@@ -0,0 +1,164 @@
...
+
+    public void addRow(RawMetricDataSet row)
+    {
+        if (_locked)
+        {
+            throw new IllegalStateException("Stats have already been locked");
+        }
+        if (!row.isIgnoreInQC() &&
+                _guideSet.getTrainingStart().compareTo(row.getAcquiredTime()) <= 0 &&
+                (_guideSet.getTrainingEnd() == null || _guideSet.getTrainingEnd().compareTo(row.getAcquiredTime()) >= 0))
```
labkey-ankurj, line 87: Could row.getAcquitedTime ever be null here?

labkey-jeckels: No, it's now filtered out in the query. A test had a case where one sample file has a null acquired time, and without any idea of where to put it in the time sequence, it's not useful for QC purposes

labkey-ankurj (approved):
Looks good.

PR #177, opened 2020-05-14
Issue 40407: Chromatogram library folder issues

vagisha:
#### Rationale
Fixes the following bugs:
1. Importing a document that has a protein with the same sequence but a different label from one that is already in the library does not cause a conflict.
2. Selecting multiple documents and queuing them for import builds a library only from the first imported document. The library is not updated after the other documents are imported.
3. Protein names in the "Current Library Protein" column in the conflict resolution UI cannot be read (white text on a grey background).

#### Changes
Updated TargetedMSLibraryTest and input files as described in https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40407


labkey-jeckels (approved):
Looks good, thanks for the test coverage!

PR #179, opened 2020-05-17
Issue 40469: java.util.zip.DataFormatException if chromatogram size i…

labkey-jeckels:
…s unknown and bytes are not compressed

#### Rationale
It's possible for the targetedms.precursorchrominfo table to have chromatograms where we don't know the uncompressed size, but the byte array hasn't been compressed. It appears that even within a single Skyline document, we can have a mix of compressed and uncompressed chromatograms.

### DIFF
--- a/src/org/labkey/targetedms/parser/SkylineBinaryParser.java
+++ b/src/org/labkey/targetedms/parser/SkylineBinaryParser.java
@@ -300,7 +300,23 @@ public class SkylineBinaryParser
             // For older data that got saved in the database without a value for uncompressedSize
             uncompressedSize = (Integer.SIZE / 8) * numPoints * (numTransitions + 1);
         }
-        return uncompress(bytes, uncompressedSize);
+        try
+        {
+            return uncompress(bytes, uncompressedSize);
+        }
+        catch (DataFormatException e)
+        {
+            // Issue 40469 - we may have -1 from a legacy import, but not have compressed data. If so, the attempt
+            // to uncompress will blow up and we can just return the original bytes
+            if (uncompressedSize.intValue() == -1)
+            {
+                return bytes;
+            }
+            else
+            {
+                throw e;
+            }
+        }
     }
 
     public static byte[] uncompress(byte[] bytes, Integer uncompressedSize) throws DataFormatException


src/org/labkey/targetedms/parser/SkylineBinaryParser.java
```diff
@@ -300,7 +300,23 @@ public int matchTransitions(ChromGroupHeaderInfo header, List<? extends GeneralT
...
-        return uncompress(bytes, uncompressedSize);
+        try
+        {
+            return uncompress(bytes, uncompressedSize);
+        }
+        catch (DataFormatException e)
+        {
+            // Issue 40469 - we may have -1 from a legacy import, but not have compressed data. If so, the attempt
+            // to uncompress will blow up and we can just return the original bytes
+            if (uncompressedSize.intValue() == -1)
```
labkey-ankurj, line 311: can we do this check before the return uncompress(bytes, uncompressedSize); ?

labkey-ankurj (approved):

src/org/labkey/targetedms/parser/SkylineBinaryParser.java
labkey-jeckels: Nope. I had hoped so too, but we don't know if it's compressed data or not.

nickshulman:
By the way,
The code in Skyline that figures out whether something is compressed is in ChromatogramGroupInfo.CompressedBytesToTimeIntensities.
If "uncompressedSize" is -1, then it calculates what the uncompressed size should be by multiplying the number of transitions, and number of time points, and a bunch of other things. Then, if that number is equal to the "compressedSize", then it knows that it's not compressed.

You might want to do the same calculation here, or I imagine that the way you are doing it catching an exception when you try to decompress works fine.

https://github.com/ProteoWizard/pwiz/blob/master/pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs#L2570

labkey-jeckels:
@nickshulman great, thanks! We do that calculation if the uncompressedSize is null, but not if it's -1. I was worried that it wouldn't always be correct. I'll keep the 20.3 change as-is, but I've reopened the issue to refine in 20.7 since it's cleaner to avoid the failed uncompress attempt

PR #192, opened 2020-06-01
Issue 40469: java.util.zip.DataFormatException if chromatogram size is -1

labkey-jeckels:
and bytes are not compressed

#### Rationale
I made a 20.3 hotfix to recover from this case where we tried to uncompress but didn't need to, but we can identify this ahead of time based on input from Nick.

#### Changes
If the size is unknown, we can calculate from other available info just like Skyline does when it's handling chromatograms


labkey-jeckels:
@labkey-ankurj I think you suggested something like this in the original hotfix PR but I didn't know for sure when it was safe to calculate the size up front. Now we do.

labkey-ankurj (approved):

PR #198, opened 2020-06-04
Check Bibliospec library source files 40051

kaipot:


api-src/org/labkey/api/targetedms/BlibSourceFiles.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public BlibSourceFiles(List<String> spectrumSourceFiles, List<String> idFiles)
+    {
+        this.spectrumSourceFiles = spectrumSourceFiles;
+        this.idFiles = idFiles;
+    }
+
+    public List<String> getSpectrumSourceFiles()
+    {
+        return spectrumSourceFiles;
```
vagisha: Maybe return Collections.unmodifiableList() here and in getIdFiles()?

vagisha (approved):
Looks OK to me, other than the comment I left about using Collections.unmodifiableList.  Wait for Josh to take a look.

labkey-jeckels (approved):
Vagisha's suggestion of using the unmodifiable wrapper sounds good but not strictly required either. Please update the PR title to reference the issue number before merging.

PR #200, opened 2020-06-10
Item 7278: Improve QC folder query perf - remove JS outlier calculations

labkey-ankurj:
#### Rationale
This work removes the javascript outlier calculations for the qc plots graph utilizing the previous refactor work done to speed up the outlier calculations in Java.

#### Changes
* new Api - GetQCPlotsData in TargetedMSController
* removal of JS outlier calculations


src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -49,19 +52,45 @@ public static OutlierGenerator get()
...
+        sql.append("\n FROM ").append(schemaName).append('.').append(queryName);
+        if (!annotationGroups.isEmpty())
+        {
+            sql.append(" WHERE ");
+            StringBuilder filterClause = new StringBuilder("SampleFileId.ReplicateId in (");
+            var intersect = "";
+            var selectSql = "(SELECT ReplicateId FROM targetedms.ReplicateAnnotation WHERE ";
+            for (AnnotationGroup annotation : annotationGroups)
+            {
+                filterClause.append(intersect).append(selectSql).append(" Name='").append(annotation.getName()).append("' AND ( ");
```
labkey-jeckels: Need to escape quotes here, since annotation names aren't encoded

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -49,19 +52,45 @@ public static OutlierGenerator get()
...
+                var or = "";
+                for (String value : annotation.getValues())
+                {
+                    filterClause.append(or).append("Value='").append(value).append("'");
```
labkey-jeckels: Same thing - need to escape value's value

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -49,19 +52,45 @@ public static OutlierGenerator get()
...
+            }
+            filterClause.append(") ");
+            sql.append(filterClause.toString());
+        }
+        sql.append(")");
+        return sql.toString();
     }
 
-    private String queryContainerSampleFileRawData(List<QCMetricConfiguration> configurations)
+    private String queryContainerSampleFileRawData(List<QCMetricConfiguration> configurations, String startDate, String endDate, List<AnnotationGroup> annotationGroups)
```
labkey-jeckels: Can these be Dates instead of Strings? Perhaps push the type back into the form that's receiving the user's input?

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -106,14 +135,26 @@ private String queryContainerSampleFileRawData(List<QCMetricConfiguration> confi
         sql.append("\nON sf.ReplicateId = exclusion.ReplicateId AND (exclusion.MetricId IS NULL OR exclusion.MetricId = x.MetricId)");
         sql.append("\nLEFT JOIN GuideSetForOutliers gs");
         sql.append("\nON ((sf.AcquiredTime >= gs.TrainingStart AND sf.AcquiredTime < gs.ReferenceEnd) OR (sf.AcquiredTime >= gs.TrainingStart AND gs.ReferenceEnd IS NULL))");
-        sql.append("\nWHERE sf.AcquiredTime IS NOT NULL");
+        if (null != startDate && null != endDate)
```
labkey-jeckels: Is there a case where we'd get a start but not an end, or vice versa? Even if the UI won't produce it at the moment, might be worth it to support either/or

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -194,4 +235,76 @@ public String getMetricLabel(Map<Integer, QCMetricConfiguration> metrics, RawMet
...
+    /**
+     * returns the separated plots data per peptide
+     * */
+    public List<QCPlotFragment> getQCPlotFragment(List<RawMetricDataSet> rawMetricData, Map<GuideSetKey, GuideSetStats> stats)
+    {
+        List<QCPlotFragment> qcPlotFragments = new ArrayList<>();
+        Map<String, List<RawMetricDataSet>> rawMetricDataSetMapByLabel = new HashMap<>();
+        for (RawMetricDataSet rawMetricDataSet : rawMetricData)
+        {
+            if (null == rawMetricDataSetMapByLabel.get(rawMetricDataSet.getSeriesLabel()))
```
labkey-jeckels: I think computeIfAbsent() or a similar method would simplify this code

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -49,19 +52,45 @@ public static OutlierGenerator get()
...
+            var intersect = "";
+            var selectSql = "(SELECT ReplicateId FROM targetedms.ReplicateAnnotation WHERE ";
+            for (AnnotationGroup annotation : annotationGroups)
+            {
+                filterClause.append(intersect).append(selectSql).append(" Name='").append(annotation.getName()).append("' AND ( ");
+                var or = "";
+                for (String value : annotation.getValues())
+                {
+                    filterClause.append(or).append("Value='").append(value).append("'");
+                    or = " OR ";
```
labkey-jeckels: Could also use an IN clause

webapp/TargetedMS/js/LeveyJenningsPlotHelper.js
```diff
@@ -12,59 +12,52 @@ Ext4.define("LABKEY.targetedms.LeveyJenningsPlotHelper", {
...
-                if (!this.guideSetDataMap[guideSetId]) {
-                    this.guideSetDataMap[guideSetId] = this.getGuideSetDataObj(row);
-                    this.hasGuideSetData = true;
-                }
+        Ext4.each(plotDataRows, function(plotDataRow) {
+            Ext4.each(plotDataRow.GuideSetStats, function(guideSetStat) {
+                var guideSetId = guideSetStat['GuideSetId'];
+                var seriesLabel = plotDataRow['SeriesLabel'];
+
+                if (guideSetId > 0) {
```
labkey-jeckels, line 25: 0 is the value we use for the default guide set - does this handle the case where not guide sets are defined?

webapp/TargetedMS/js/LeveyJenningsPlotHelper.js
```diff
@@ -158,7 +151,7 @@ Ext4.define("LABKEY.targetedms.LeveyJenningsPlotHelper", {
     {
         var data = {};
         // if a guideSetId is defined for this row, include the guide set stats values in the data object
-        if (Ext4.isDefined(row['GuideSetId']))
+        if (Ext4.isDefined(row['GuideSetId']) && row['GuideSetId'] > 0)
```
labkey-jeckels, line 155: Same question here about the default guide set behavior

webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -126,125 +69,108 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
...
-                            MeanMR: guideSetAvgMRs[guideSetId].Series[seriesLabel][seriesType].avgMR,
-                            StdDevMR: guideSetAvgMRs[guideSetId].Series[seriesLabel][seriesType].stddevMR
+                            MeanMR: guideSetStat['MRMean'],
+                            StdDevMR: guideSetStat['MRStdDev']
                         };
                     }
                     else {
-                        this.guideSetDataMap[guideSetId].Series[seriesLabel][seriesType].MeanMR = guideSetAvgMRs[guideSetId].Series[seriesLabel][seriesType].avgMR;
-                        this.guideSetDataMap[guideSetId].Series[seriesLabel][seriesType].StdDevMR = guideSetAvgMRs[guideSetId].Series[seriesLabel][seriesType].stddevMR;
+                        this.guideSetDataMap[guideSetId].Series[seriesLabel][seriesType].MeanMR = guideSetStat['MRMean'];
```
labkey-jeckels: Unless there's a compelling reason to use different names, can we make these consistent?

labkey-jeckels (changes requested):
Looking good - a few ideas and questions.

webapp/TargetedMS/js/LeveyJenningsPlotHelper.js
labkey-ankurj: Yes, there are two different objects for both cases defaultGuideSet when no guidesets are defined and guideSetMap when guideset is defined

src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -1057,8 +1059,8 @@ public Object execute(QCMetricOutliersForm form, BindException errors)
         private boolean _includeMR;
         private boolean _includeMeanCusum;
         private boolean _includeVariableCusum;
-        private String _startDate;
-        private String _endDate;
+        private java.sql.Date _startDate;
```
labkey-jeckels: Let's use java.util.Date instead - hopefully an easy swap

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -64,14 +64,31 @@ private String getEachSeriesTypePlotDataSql(int seriesIndex, int id, String sche
...
             for (AnnotationGroup annotation : annotationGroups)
             {
-                filterClause.append(intersect).append(selectSql).append(" Name='").append(annotation.getName()).append("' AND ( ");
-                var or = "";
-                for (String value : annotation.getValues())
+                filterClause.append(intersect)
+                        .append(selectSql)
+                        .append(" Name=")
+                        .append(escapeQuote)
+                        .append(annotation.getName())
```
labkey-jeckels: The concern is that the annotation name or value will have a quote in it. Need to something like:

.append(" Name='").append(annotation.getName().replace("'", "''")).append("'");

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -64,14 +64,31 @@ private String getEachSeriesTypePlotDataSql(int seriesIndex, int id, String sche
...
+                        .append(escapeQuote);
+
+                var annotationValues = annotation.getValues();
+                if (!annotationValues.isEmpty())
                 {
-                    filterClause.append(or).append("Value='").append(value).append("'");
-                    or = " OR ";
+                    filterClause.append(" AND  Value IN (")
+                            .append(escapeQuote)
+                            .append(annotationValues.get(0))
```
labkey-jeckels: Suggest using StringUtils.join() to build up the quote-escaped, comma-separated list

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -135,14 +152,27 @@ private String queryContainerSampleFileRawData(List<QCMetricConfiguration> confi
...
+
+            if (null != endDate)
+            {
+                sql.append("\n sf.AcquiredTime < TIMESTAMPADD('SQL_TSI_DAY', 1, CAST('");
+                sql.append(endDate);
+                sql.append("' AS TIMESTAMP))");
+            }
+            else
+            {
+                sql.append("1=1");
```
labkey-jeckels: Perhaps cleaner to use a separator variable that's first set to "WHERE" and then to "AND", to avoid the 1=1 in the else

labkey-jeckels (changes requested):
Some followups - sorry for not giving more detailed guidance before.

I also see that the tests are sad, maybe because of the quote escaping changes?

src/org/labkey/targetedms/TargetedMSController.java
labkey-ankurj: not able to directly use java.util.Date in sql query.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-ankurj: I guess you meant this: 
`var quoteEscapedVals = annotationValues.stream().map(s -> s.replace("'", "''")).collect(Collectors.toList());
filterClause.append(" AND  Value IN (").append(StringUtils.join(quoteEscapedVals, ","));`

still a for loop would be required to build quotes around the strings?

src/org/labkey/targetedms/TargetedMSController.java
labkey-ankurj: well no, it seems to be working, something was not right initially when I was trying to use it.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-ankurj: done

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-ankurj: got it, changed

src/org/labkey/targetedms/TargetedMSController.java
labkey-ankurj: replaced

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -64,31 +65,21 @@ private String getEachSeriesTypePlotDataSql(int seriesIndex, int id, String sche
...
-                    for (int i = 1; i < annotationValues.size(); i++)
-                    {
-                        filterClause.append(",")
-                                .append(escapeQuote)
-                                .append(annotationValues.get(i))
-                                .append(escapeQuote);
-                    }
+                    var quoteEscapedVals = annotationValues.stream().map(s -> s.replace("'", "''")).collect(Collectors.toList());
+                    var vals = "'" + StringUtils.join(quoteEscapedVals, "','") + "'";
+                    filterClause.append(" AND  Value IN (").append(vals);
```
labkey-jeckels: I haven't tried running this, but the parentheses seems suspicious. Shouldn't there be a closing paren inside of the if() statement's body?

labkey-jeckels (approved):
See one comment on parentheses, and subject to happy tests on TeamCity

PR #204, opened 2020-06-16
Support for crosslinked peptides in Skyline.

nickshulman:
#### Rationale
Support for new Skyline document format related to crosslinked peptides
Skyline 20.2 will support crosslinked peptides.
Crosslinked peptides are represented as explicit modifications in the XML.
In addition, the Transitions have "linked_fragment_ion" elements under them which fragments from the linked peptides are attached to the main fragment ion.

#### Related Pull Requests
* 

#### Changes
The class "ComplexFragmentIonName" gets constructed from the "linked_fragment_ion" elements under the <transition> element. The toString() of this object gets stored in the "complexFragmentIon" column in the Transition table. This string gets used by PeakAreaRatioCalculator.getPeptideTransitionKey to make it so that transitions are unique within a particular precursor.


labkey-jeckels:
Thanks Nick! I just pushed a SQL Server version of the script to make the tests happier.

Is there really just a single column change to the DB to handle all of this? I was expecting more churn on the schema based on the diffs I generated from the Skyline XSD changes

nickshulman:
So far, I think we can get away with just adding a single column.

Skyline represents the crosslinked structure in the .sky file with
"<linked_peptide>" elements under the "<explicit_modifications>" element.
Also, Skyline represents fragment ions with "<linked_fragment_ion>"
elements.

Panorama does not really care about being able to parse the structure of
the crosslinks, so it can ignore everything under
"<explicit_modifications>". Then, I have the class "ComplexFragmentIonName"
which takes all of the <linked_fragment_ion> elements and turns them into a
flat string. That string gets stored in the new column
"targetedms.transition.complexFragmentIon", and it gets used as part of the
key in PeakAreaRatioCalculator.getPeptideTransitionKey

In the future, we might want to make it so that LabKey is able to parse the
complexFragmentIon string, so that transitions can be displayed better.
For instance, the value "p-{6:DSS:b4}" is the precursor from the first
peptide combined with the b4 ion from the peptide that's attached to the
"DSS" crosslinker at position 6 on the first peptide.
In Skyline, when it notices that there are only two peptides crosslinked to
each other, that transition would just be displayed as "p-b4". In order to
be able to do this right, we would probably need to start parsing the
<explicit_modifications> element in the .sky file in order to know how
complicated the structure of the crosslinks is.
-- Nick

labkey-jeckels:
OK, cool.

I see you added some files and make sure that they get imported in a test, which is great. I'm assuming we should be adding a little more validation that the modifications show up in the UI, but will have to take a look at the results of the import to understand the current display.

labkey-jeckels:
@labkey-sweta can you review my test changes here? I've done the review of the product-side changes, so if the test looks good to you, we can merge

test/src/org/labkey/test/tests/targetedms/TargetedMSDocumentFormatsTest.java
```diff
@@ -14,39 +14,121 @@
...
-        }
+        goToProjectHome(getProjectName());
+        importData(SAMPLEDATA_FOLDER + "DocumentSerializerTest_36.sky.zip", ++JOB_COUNT);
+    }
+
+    @Test
+    public void testNormalDocument()
+    {
+        goToProjectHome(getProjectName());
+        importData(SAMPLEDATA_FOLDER + "DocumentSerializerTest.sky.zip", ++JOB_COUNT);
```
labkey-sweta, line 60: No verification needed?

test/src/org/labkey/test/tests/targetedms/TargetedMSDocumentFormatsTest.java
```diff
@@ -14,39 +14,121 @@
...
     }
 
-    private void testImportDoc(String filename) {
-        setupSubfolder(getProjectName(), filename, FolderType.Experiment);
-        importData(SAMPLEDATA_FOLDER + filename + ".sky.zip");
+    @Test
+    public void testCompactDocument()
```
labkey-sweta, line 64: Same question, just importing with no verification.

labkey-sweta (approved):

test/src/org/labkey/test/tests/targetedms/TargetedMSDocumentFormatsTest.java
labkey-jeckels: That's right. The original version of this test just imported three files and verified there were no errors. This branch has added two new files with cross-linked peptide data, and those are the ones where I did more extensive validation.

PR #205, opened 2020-06-17
Issue 40235: Improve reporting on targetedms.AuditLogMessage and AuditLogEntry

labkey-jeckels:
Push CTE into a DB view to make it easy to always pull the ID of the run associated with the entry


labkey-jeckels:
I backported the 20.000-20.001 change to trigger a schema version change in 20.3 to get the targetedms-create.sql script to run so that the view gets created. When we merge this forward, we'll need to bump TargetedMSModule.getSchemaVersion() up one (20.007 at the moment) to make sure that dev machines also get the view created.

labkey-ankurj (approved):

PR #208, opened 2020-06-17
Issue 38963: Incorrect precursor plots in small molecule QC folder

labkey-ankurj:
#### Rationale
This work centralizes the calculation of series label for precursors in OutlierGenerator.queryContainerSampleFileRawData and removes the series label from individual metrics query. The results of which correct the number of precursor plots in a small molecule QC.
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38963

#### Related PR
https://github.com/LabKey/targetedms/pull/191


webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1146,66 +1146,36 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
-    setPrecursorsForPage: function() {
+    setPrecursorsForPage: function(plotDataRows) {
         if (Ext4.isNumeric(LABKEY.ActionURL.getParameter('qcPlots.offset')))
             this.pagingStartIndex = parseInt(LABKEY.ActionURL.getParameter('qcPlots.offset'));
 
         if (this.pagingStartIndex < 0)
             this.pagingStartIndex = 0;
-        else if (this.pagingStartIndex > this.allPrecursors.length)
-            this.pagingStartIndex = this.allPrecursors.length - this.maxCount;
+        else if (this.pagingStartIndex > plotDataRows.length)
```
labkey-jeckels: This doesn't feel right. Is this already organized by precursor? We want to show all the data for 20 (or whatever the cap is) precursors per page, not just take the first X rows that came back.

labkey-jeckels (changes requested):
Maybe I'm misreading the code, but it feel off to me.

webapp/TargetedMS/js/QCTrendPlotPanel.js
labkey-ankurj: plotDataRows is the list of QCPlotFragment and it is organized by precursor.

labkey-jeckels: So if you edit QCTrendPlotPanel's maxCount to be 2, you see all the data for the first two precursors, and can go to the next set?  Assuming that works, maybe the variable can be renamed to plotDataRowsByPrecursor or something that implies that it's not just a flat array of all the data rows?

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -102,14 +102,14 @@ private String queryContainerSampleFileRawData(List<QCMetricConfiguration> confi
         sql.append("\nCOALESCE(pci.PrecursorId.Id, pci.MoleculePrecursorId.Id) AS PrecursorId,");
 
         sql.append("\nCOALESCE(X.SeriesLabel, COALESCE(pci.PrecursorId.ModifiedSequence,");
```
labkey-jeckels: Can we just select X.SeriesLabel here, and then do a null check in getSeriesLabel() to lazily create the string as needed? And also eliminate the need to call buildSeriesLabel() separately?

labkey-jeckels (changes requested):
I think we can switch to always generate the label (when not present in the underlying QC query) in Java, instead of doing this split SQL/Java approach.

src/org/labkey/targetedms/model/RawMetricDataSet.java
```diff
@@ -47,14 +47,61 @@
...
     public String getSeriesLabel()
     {
-        return seriesLabel;
+        if (null != seriesLabel)
+        {
+            return seriesLabel;
+        }
+        else if (null != modifiedSequence)
+        {
+            return modifiedSequence;
```
labkey-jeckels: Overall structure looks as I'd expect, but we can't just use modifiedSequence - we need to include the mZ and charge state - please double-check against the SQL that Vagisha proposed in the issue.

labkey-jeckels (changes requested):

src/org/labkey/targetedms/model/RawMetricDataSet.java
```diff
@@ -61,42 +61,48 @@ public String getSeriesLabel()
...
+                    modifiedSL.append(" [");
+                    modifiedSL.append(Math.round(massMonoisotopic * scale) / scale);
+                    modifiedSL.append("/");
+                    modifiedSL.append(Math.round(massAverage * scale) / scale);
+                    modifiedSL.append("] ");
+                }
             }
 
-            if (null != this.massMonoisotopic && null != this.massAverage)
+            if (null != massAverage)
```
labkey-jeckels: Shouldn't this be mz instead of massAverage?

And let's use DecimalFormat instead of Math.round(), which will probably have some of the same floating point problems. Sorry for not catching this last aspect earlier.

labkey-jeckels (changes requested):
Final round, I hope

labkey-jeckels (approved):
Code looks good now, please merge when tests have run successfully

labkey-jeckels (approved):
Please merge when tests are looking good

test/src/org/labkey/test/tests/targetedms/TargetedMSQCGuideSetTest.java
```diff
@@ -480,7 +494,7 @@ private void verifyGuideSet5Stats(GuideSet gs) throws IOException, CommandExcept
 
     private void verifyGuideSetSmallMoleculeStats(GuideSet gs) throws IOException, CommandException
     {
-        String precursor = "C16";
+        String precursor = "C16,";
```
labkey-ankurj, line 497: @labkey-jeckels had to do this to filter out series labels having C16:0, C16:1, C16-H ... Let me know if you have another suggestion

PR #209, opened 2020-06-18
Issue 40659: Security fixes for vulnerabilities found in security scan report

labkey-ankurj:
#### Rationale
This PR updates the jQuery version used in targetedms module.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1317


labkey-jeckels (approved):
This file doesn't appear to be used, so safe to delete. May still need followup changes to make sure spectrum viewer and the precursor resolution pages keep working when the core flips to 3.5.x

labkey-jeckels (changes requested):
How does the code know to use the new file name? And do we need a separate copy of these files - could we reference the ones in the platform repo?

labkey-ankurj:
> How does the code know to use the new file name? And do we need a separate copy of these files - could we reference the ones in the platform repo?

Right, jQuery is not being used.

labkey-jeckels (approved):

PR #210, opened 2020-06-18
Issue 38724: Panorama QC plots can have y-axis label overlaying values

labkey-ankurj:
#### Rationale
Dynamically add space between yAxis values and yAxis label in QC plots so that they don't overlap.
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=38724


labkey-jeckels:
No complaints here, but I imagine Cory will be better qualified to evaluate

webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -627,11 +627,13 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
...
         var mainTitle = LABKEY.targetedms.QCPlotHelperWrapper.getQCPlotTypeLabel(plotType, isCUSUMMean);
 
+        var leftMargin = 75;
+
         var basePlotConfig = this.getBasePlotConfig(id, precursorInfo.data, plotLegendData);
         var plotConfig = Ext4.apply(basePlotConfig, {
             margins : {
                 top: 65 + this.getMaxStackedAnnotations() * 12,
-                left: 75,
+                left: leftMargin + this.getYAxisLeftMarginOffset(precursorInfo),
```
cnathe: It is a rather quick calculation, but we are doing `this.getYAxisLeftMarginOffset(precursorInfo)` three times. Instead, we could just move that up to a variable at line 631 and then use it here and in the position value below.

webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -645,7 +647,8 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
                 },
                 yLeft: {
                     value: this.getYScaleLabel(plotType, trendLineProps.valueConversion, metricProps.series1Label),
-                    color: this.isMultiSeries() ? this.getColorRange()[0] : undefined
+                    color: this.isMultiSeries() ? this.getColorRange()[0] : undefined,
+                    position: this.getYAxisPosition(precursorInfo, leftMargin)
```
cnathe: If you move `this.getYAxisLeftMarginOffset(precursorInfo)` to a variable above, then `getYAxisPosition` can probably go away and just be inlined here.
```suggestion
                    position: marginOffset > 0 ? leftMargin + marginOffset - 15 : undefined
```

webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -688,6 +691,33 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
...
+                undefined
+    },
+
+    getYAxisLeftMarginOffset: function(precursorInfo)
+    {
+        var maxLength = precursorInfo.min.toString().length;
+
+        if (maxLength < precursorInfo.max.toString().length) {
+            maxLength = precursorInfo.max.toString().length;
+        }
```
cnathe: I think this can be simplified to a single line using `Math.max()`
```suggestion
        var maxLength = Math.max(precursorInfo.min.toString().length, precursorInfo.max.toString().length)
```

cnathe: Also, I assume there is no reason that the`precursorInfo.min` or `precursorInfo.max` would ever be undefined, but might be worth a quick check.
```
if (precursorInfo.min === undefined || precursorInfo.max === undefined) {
    return 0;
}  
```

cnathe (changes requested):

webapp/TargetedMS/js/QCPlotHelperBase.js
labkey-ankurj: done

webapp/TargetedMS/js/QCPlotHelperBase.js
labkey-ankurj: done

webapp/TargetedMS/js/QCPlotHelperBase.js
labkey-ankurj: done

cnathe (approved):

PR #214, opened 2020-06-24
Improve incremental rendering layout for QC Summary web part

labkey-jeckels:
#### Rationale
The QC summary web part is resizing at least twice during a standard render as it gets API responses. That's a lot of flashing for the user, and it's causing test failures too as the content shifts while the test is interacting with it

#### Changes
Reserve the expected amount of space for the web part when it has no child folders, and when it does, reserve enough space for them as soon as we know how many there will be. When there's < 3 samples loaded in a folder (uncommon) we'll reserve a little extra room more than we'll need, but it won't be needed elsewhere with so little data, so no problem.

### DIFF
--- a/src/org/labkey/targetedms/view/qcSummary.jsp
+++ b/src/org/labkey/targetedms/view/qcSummary.jsp
@@ -38,7 +38,7 @@
     Integer sampleLimit = (Integer)HttpView.currentModel();
 %>
 
-<div id=<%=q(qcSummaryId)%>></div>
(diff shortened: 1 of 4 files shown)


labkey-ankurj (approved):
QC Summary part behaves better now.

labkey-jeckels:
Looks like this doesn't fully resolve the problem with the popup in QC folders, but it's still worth taking as it's an improved user experience.

PR #220, opened 2020-07-06
Issue 40831: Overflowing integer values on PK for targetedms.TransitionChromInfo

labkey-jeckels:
#### Rationale
PanoramaWeb.org has overflowed the integer (4-byte) values for the targetedms.TransitionChromInfo table's Id column.

#### Changes
Both the Java code and the DB tables are using 4-byte ints and will need to switch to long/bigint. While we should widen all or many of the PKs in the targetedms schema, this is the only table that's close (targetedms.precursorchominfo is at ~330 million, well short of 2 billion). Thus, we can just handle the code update in 20.3 and run SQL against the one affected server to unblock, and do 20.11 work to make all of the tables support the wider column.

### DIFF
--- a/src/org/labkey/targetedms/SkylineDocImporter.java
+++ b/src/org/labkey/targetedms/SkylineDocImporter.java
@@ -1708,7 +1708,7 @@ public class SkylineDocImporter
     {
         for (TransitionChromInfo transChromInfo : transitionChromInfos)
         {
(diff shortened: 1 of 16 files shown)


labkey-adam (approved):

labkey-ankurj (approved):

vagisha (approved):
Thanks!

PR #233, opened 2020-07-18
Test updates due to update to GetQueryDetailsAction

labkey-martyp:
#### Rationale
GetQueryDetailsAction was updated to return the public name instead of the name that was passed in.  That public name is displayed in the details section of the schema browser.  These test updates account for that change.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1379

#### Changes
- Various test updates to look for public name to verify details have loaded

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSSchema.java
+++ b/src/org/labkey/targetedms/TargetedMSSchema.java
@@ -83,6 +83,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -1439,7 +1440,7 @@ public class TargetedMSSchema extends UserSchema
     @Override
     public Set<String> getTableNames()
     {
-        CaseInsensitiveHashSet hs = new CaseInsensitiveHashSet();
+        HashSet hs = new HashSet();
         hs.add(TABLE_TARGETED_MS_RUNS);
         hs.add(TABLE_ENZYME);
         hs.add(TABLE_RUN_ENZYME);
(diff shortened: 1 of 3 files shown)


labkey-jeckels:
I understand needing to update tests based on the server change, but is there any way we can avoid needing to pass in two different casings for these table names? Do we need to use the public name in more places, like the list of tables? Or do a case-insensitive match somewhere in the test code?

labkey-martyp:
@labkey-jeckels Yeah I had done that originally with a case-insensitive check but it has issues when we do the special character names.  I can probably update that to try case insensitive match then case sensitive if that doesn't work.
https://github.com/LabKey/testAutomation/blob/85c8115b00772558a021e18ce18fcbceb702b6f4/src/org/labkey/test/Locator.java#L1210

labkey-jeckels:
Are we showing one variant in the tree, and the other in the details pane? It would be nice for the server to return a single, canonical name, and a not a mix of names

labkey-martyp:
Yeah looks like we're using getTitle() there instead of getPublicName().  Think I should use that in the GetQueryDetailsAction?
https://github.com/LabKey/platform/blob/b19a174d884bef41a7c95fc2008a6b1f3651520b/query/src/org/labkey/query/controllers/GetSchemaQueryTreeAction.java#L149

labkey-jeckels:
I think this should keep calling getTitle(), but that we should update the implementation of that method.

getTitle() allows for a query to have a different label from its real name. For example, in the EHR, the "drug" dataset has a label of "Drug Administration". In the absence of a set label, though, I'd say getTitle() should return the public table name.

@labkey-matthewb do you agree?

labkey-martyp:
Yeah looks like getTitle returns name if there is no title.  To be clear, the tree in the schema browser displays "name (title)" for the tables/queries if name is not equal to title, otherwise just name.  The details section of schema browser displays just the name of the query as returned from GetQueryDetailsAction. Right now GetQueryDetailsAction is setting name to whatever is passed in, instead of what is saved on the server as name, title or public name.  So for instance in biologics it could be ExpressionSystemSamples or expressionsystemsamples, depending on what a user might type into the url.  And then we have unpredictable results, such as when we pass it in as part of an IN filter.

labkey-martyp:
Sorry correction on that last comment, the details section of schema browser is displaying "name (title)" if they are not the same.

I'll try just returning the query name as name from GetQueryDetailsAction, but my initial testing with that had issues with some user schema queries that had different underlying db schema table names.

PR #234, opened 2020-07-23
Issue 40959: Unable to resolve chromatograms after Skyline centroiding change

labkey-jeckels:
#### Rationale
Skyline used to store a different path for the centroided version of a sample file, of the form:

T:\path\to\file.mzML?centroid_ms1=true

It no longer uses the query parameter section this way, so we have to check for the presence of absence of the parameter when matching up paths

### DIFF
--- a/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
+++ b/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
@@ -61,6 +61,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -2719,13 +2763,29 @@ private static int compareMz(SignedMz precursorMz1, SignedMz precursorMz2, doubl
...
+            // Sample-scoped chromatograms have a magic precursor MZ value
             if (chromatogram.getPrecursorMz() == 0.0)
             {
                 String path = _binaryParser.getFilePath(chromatogram);
                 SampleFile sampleFile = pathToSampleFile.get(path);
+
+                // Issue 40959 - Try stripping off the URI parameters if we don't have a match
+                if (sampleFile == null)
+                {
+                    int queryParamIdx = path.indexOf('?');
```
vagisha: Use lastIndexOf('?') here as well?

vagisha (approved):

labkey-ankurj (approved):

PR #237, opened 2020-07-28
Issue 40998: Only show pivot columns in Panorama sample-based reports…

labkey-jeckels:
… for sample files that match the current run

#### Rationale
Users are interested in custom Panorama reports that pivot data into per-sample result columns. They want to view this for just the samples in a single Skyline file, and our normal PIVOT behavior will create columns for all of the samples in the current container.

#### Related PRs
https://github.com/LabKey/targetedms/pull/41

#### Changes
Restore a slightly more generalized version of a customizer that limits the columns based on the run id filter on the URL

### DIFF
--- a/src/org/labkey/targetedms/query/SamplePivotCustomizer.java
+++ b/src/org/labkey/targetedms/query/SamplePivotCustomizer.java
@@ -1,0 +1,89 @@
+package org.labkey.targetedms.query;
+
+import org.apache.commons.collections4.MultiValuedMap;
+import org.labkey.api.collections.CaseInsensitiveHashSet;
+import org.labkey.api.data.CompareType;
+import org.labkey.api.data.SimpleFilter;
+import org.labkey.api.data.TableCustomizer;
+import org.labkey.api.data.TableInfo;
+import org.labkey.api.data.TableSelector;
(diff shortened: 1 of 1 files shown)


src/org/labkey/targetedms/query/SamplePivotCustomizer.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Customizes the set of columns available on a pivot query that operates on samples, hiding all of the pivot values
+ * that aren't part of the run that's being filtered on. This lets you view a single document's worth of data
+ * without seeing empty columns for all of the other samples in the same container.
+ */
+public class SamplePivotCustomizer implements TableCustomizer
```
labkey-ankurj, line 27: Is this supposed to be referenced from some query.xml?

labkey-ankurj (approved):

src/org/labkey/targetedms/query/SamplePivotCustomizer.java
labkey-jeckels: Yes, it's a custom report on a customer server

PR #243, opened 2020-08-16
Item 7541: Convert ints to bigints/long for targetedms tables to avoid overflowing integer values

labkey-ankurj:
#### Rationale
This work is continuation of Overflowing integer Issue in the related pull request.

#### Related Pull Requests
https://github.com/LabKey/targetedms/pull/220

### DIFF
--- a/src/org/labkey/targetedms/chromlib/LibPeptideStructuralModification.java
+++ b/src/org/labkey/targetedms/chromlib/LibPeptideStructuralModification.java
@@ -98,8 +98,8 @@ public class LibPeptideStructuralModification implements ObjectWithId
     @Override
     public int hashCode()
     {
-        int result = Math.toIntExact(_peptideId);
-        result = Math.toIntExact(31 * result + _structuralModificationId);
+        int result = (int) _peptideId;
+        result = (int) (31 * result + _structuralModificationId);
         result = 31 * result + _indexAa.hashCode();
         result = 31 * result + _massDiff.hashCode();
         return result;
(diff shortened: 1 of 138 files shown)


src/org/labkey/targetedms/chromlib/LibPeptideStructuralModification.java
```diff
@@ -98,8 +98,8 @@ public boolean equals(Object o)
     @Override
     public int hashCode()
     {
-        int result = _peptideId;
-        result = 31 * result + _structuralModificationId;
+        int result = Math.toIntExact(_peptideId);
```
labkey-jeckels: This will blow up with a ArithmeticException if the ID value exceeds MAX_INT. Please double-check other new uses of Math.toIntExact as well.

For hash code purposes, I think we can just simply cast to an int - it's fine if it overflows.

src/org/labkey/targetedms/view/spectrum/LibrarySpectrumMatchGetter.java
```diff
@@ -320,7 +320,7 @@ public boolean equals(Object o)
         public int hashCode()
         {
             int result = _modifiedSequence.hashCode();
-            result = 31 * result + _generalMoleculeId;
+            result = Math.toIntExact(31 * result + _generalMoleculeId);
```
labkey-jeckels: Another example to clean up (I didn't flag most of them)

labkey-jeckels (changes requested):
Looks good, except for uses of Math.toIntExact()

I've been thinking of ways to validate the SQL upgrade script. I think we should be able to use the DB tools to script the schema's creation before and after the upgrade, and validate that the only differences in the scripts it generates are the column types. Can you try doing something like this? I want to ensure we don't lose a FK or Index accidentally

labkey-jeckels (approved):

PR #245, opened 2020-09-01
no trailing semis please

labkey-matthewb:
#### Rationale
Fix error due to use of unexpected char.  Note ';' is now a SQL token (instead of string), which apparently caused a change in behavior.  But ';' has never been part of SQL syntax.  We don't support multiple statements, so no statement terminator.

#### Related
https://github.com/LabKey/sampleManagement/pull/355


labkey-ankurj (approved):
Thanks for catching, this sql got merged from MacCossLabModules custom passport module into targetedms.

PR #252, opened 2020-09-15
Issue 41317: conditionalize sql upgrade script to support older pg vesions

labkey-ankurj:
#### Rationale
See https://github.com/LabKey/targetedms/pull/246


labkey-tchad (approved):

labkey-tchad (approved):

labkey-jeckels:
The Clustergrammer fixes are in develop already. I don't think it's worth backporting to 20.9 since the groups taking a 20.9 build aren't doing TargetedMS

labkey-tchad:
> The Clustergrammer fixes are in develop already. I don't think it's worth backporting to 20.9 since the groups taking a 20.9 build aren't doing TargetedMS

They might not be using it but that the upgrade script will still run and fail. Targetedms is in most distributions.

labkey-jeckels:
Yes, we still need to merge this change. I just meant we needn't worry about the TargetedMS failures that are being reported in this PR.

labkey-ankurj:
thanks, merging this branch.

PR #265, opened 2020-10-19
Issue 41610: Simplify ContainerJoinType.PrecursorChromInfoFK

vagisha:
#### Rationale
ContainerJoinType.PrecursorChromInfoFK does not need joins to the targetedms.runs table for container filtering since the PrecursorChromInfo table has a "Container" column.


labkey-jeckels (approved):
Looks good, assuming tests are happy after they complete

PR #268, opened 2020-10-21
Issue 40456: Make it harder for users to ignore conflicts in chromatogram library folders

vagisha:
#### Rationale
It is too easy for users to ignore conflicts in library folders.  

#### Changes
- Document upload in a library folder fails if the folder has conflicts
- If there are conflicts in a folder show the download link for the last stable library revision

### DIFF
--- a/src/org/labkey/targetedms/view/chromatogramLibraryDownload.jsp
+++ b/src/org/labkey/targetedms/view/chromatogramLibraryDownload.jsp
@@ -28,51 +28,99 @@
 <%@ page import="java.nio.file.Files" %>
 <%@ page import="java.nio.file.Path" %>
 <%@ page import="java.text.DecimalFormat" %>
+<%@ page import="org.labkey.api.security.permissions.InsertPermission" %>
+<%@ page import="org.labkey.api.util.StringUtilsLabKey" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%
     Container c = getContainer();
     User user = getUser();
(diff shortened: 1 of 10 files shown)


src/org/labkey/targetedms/view/chromatogramLibraryDownload.jsp
```diff
@@ -28,51 +28,98 @@
...
-    long peptideCount = TargetedMSController.getNumRepresentativePeptides(c);
-    long transitionCount = TargetedMSController.getNumRankedTransitions(c);
-    DecimalFormat format = new DecimalFormat("#,###");
+    long conflictCount = ConflictResultsManager.getConflictCount(user, c);
+    boolean hasConflicts = conflictCount != 0;
     int currentRevision = ChromatogramLibraryUtils.getCurrentRevision(c, user);
-    Path archiveFile = ChromatogramLibraryUtils.getChromLibFile(c, currentRevision);
+    int libRevision = hasConflicts ? ChromatogramLibraryUtils.getLastStableLibRevision(c, user) : currentRevision;
+
+    TargetedMSController.ChromLibAnalyteCounts chromLibAnalyteCounts = TargetedMSController.getChromLibAnalyteCounts(user, c, libRevision);
```
labkey-jeckels, line 44: Consider returning a ChromLibAnalyteCounts with zeros instead of a null to make this code a little cleaner. Could add an exists() method if other cases need to distinguish between null and a library with no data

labkey-jeckels (approved):
One suggestion but no additional review needed.

PR #271, opened 2020-10-24
Issue 41617: Chromatograms are not displayed for precursors that do not have any quantitative transitions

vagisha:
#### Rationale
If a precursor does not have any quantitative transitions the chromatogram chart does not display any peaks.

#### Changes
- Render non-quantitative precursor and transition peaks as dashed lines
- Use the start and end times for a precursor's transition peaks if the minStartTime and minEndTime on the precursor peak are not set


src/org/labkey/targetedms/chart/ChromatogramDataset.java
```diff
@@ -1155,9 +1206,8 @@ public int compare(TransChromInfoPlusTransition t1, TransChromInfoPlusTransition
         }
     }
 
-    static class MoleculeTransChromInfoPlusTransition
+    static class MoleculeTransChromInfoPlusTransition extends TransitionChromInfoPlusGeneralTransition<MoleculeTransition>
     {
-        private TransitionChromInfo _transChromInfo;
         private MoleculeTransition _transition;
```
labkey-jeckels: Could push this member variable up to the abstract superclass and make getTransition() non-abstract there.

src/org/labkey/targetedms/query/PrecursorManager.java
```diff
@@ -809,6 +821,44 @@ public static boolean hasOptimizationPeaks(long peptideId)
...
+        if(pci == null)
+        {
+            return null;
+        }
+        if(pci.getMaxEndTime() != null)
+        {
+            return pci.getMaxEndTime();
+        }
+
+        // Issue 41617: Chtomatograms are not displayed for precursors that do not have any quantitative transitions
```
labkey-jeckels:
```suggestion
        // Issue 41617: Chromatograms are not displayed for precursors that do not have any quantitative transitions
```

src/org/labkey/targetedms/query/PrecursorManager.java
```diff
@@ -809,6 +821,44 @@ public static boolean hasOptimizationPeaks(long peptideId)
...
+        if(pci == null)
+        {
+            return null;
+        }
+        if(pci.getMinStartTime() != null)
+        {
+            return pci.getMinStartTime();
+        }
+
+        // Issue 41617: Chtomatograms are not displayed for precursors that do not have any quantitative transitions
```
labkey-jeckels:
```suggestion
        // Issue 41617: Chromatograms are not displayed for precursors that do not have any quantitative transitions
```

labkey-jeckels (approved):
A couple of small items to consider, but up to you.

A bigger change would be to combine the queries that get the min and max values for the retention times. It looks like we always want both of them, and it would be faster to do a single query. But I don't think this will be a significant issue in practice because we're not likely to need the min/max for a lot of different transitions or precursors, correct?

vagisha:
Queries to get the min and max retention times from the transitionchrominfo table have been combined into a single query.

labkey-jeckels (approved):
Changes look good!

PR #272, opened 2020-10-27
Small molecule chromatogram library support

labkey-jeckels:
#### Rationale
We initially supported just proteomics (proteins and peptides) chromatogram libraries. We want to support small molecule use cases too.

#### Changes
* Rearrange web parts into separate portal tabs to reduce the number of large grids on one page
* Create a new version of .clib files with schema additions for small molecule data
* Report and resolve conflicts for small molecules too


resources/queries/targetedms/librarymoleculeprecursor/LibraryPrecursors.qview.xml
```diff
@@ -0,0 +1,28 @@
...
+        <column name="DeclusteringPotential" />
+        <column name="NoteAnnotations" />
+    </columns>
+
+    <filters>
+        <filter column="RepresentativeDataState" operator="eq" value="1"/>
+    </filters>
+
+    <sorts>
+        <sort column="IonFormula" />
```
vagisha, line 24: Can we sort on the MoleculeId/MoleculeName here instead of IonFormula and CustomIonName both of which can be null?

src/org/labkey/targetedms/query/ConflictResultsManager.java
```diff
@@ -88,67 +87,76 @@ private ConflictResultsManager() {}
...
-        ConflictProtein[] pepGroups = new SqlSelector(TargetedMSManager.getSchema(), getConflictProteinsSql).getArray(ConflictProtein.class);
-        return Arrays.asList(pepGroups);
+        return new SqlSelector(TargetedMSManager.getSchema(), getConflictProteinsSql).getArrayList(ConflictProtein.class);
     }
 
     public static List<ConflictPrecursor> getConflictedPrecursors(Container container)
+    {
+        List<ConflictPrecursor> result = new ArrayList<>();
+        result.addAll(getConflictedPrecursors(container, TargetedMSManager.getTableInfoPeptide(), TargetedMSManager.getTableInfoPrecursor(), "ModifiedSequence", new SQLFragment("prec.ModifiedSequence"),true));
+        result.addAll(getConflictedPrecursors(container, TargetedMSManager.getTableInfoMolecule(), TargetedMSManager.getTableInfoMoleculePrecursor(), "IonFormula", new SQLFragment("COALESCE(prec.CustomIonName, prec.IonFormula)"), false));
```
vagisha: Prefer prec.IonFormula over prec.CustomIonName in the COALESCE since precursors of a molecule will have the same CustomIonName.  Both IonFormula and CustomIonName are nullable, so maybe add GeneralPrecursor.mz here as it is done in molExprCol in MoleculeTableInfo?

src/org/labkey/targetedms/query/ConflictResultsManager.java
```diff
@@ -88,67 +87,76 @@ private ConflictResultsManager() {}
...
+        sql.append("targetedms.generalprecursor gp ");
+        sql.append("ON m.Id = gp.GeneralMoleculeId ");
+        sql.append("INNER JOIN ");
+        sql.append(precursorTable, "prec");
+        sql.append(" ON gp.Id = prec.Id ");
+        sql.append("INNER JOIN ");
+        sql.append(precursorTable, "prec2");
+        sql.append(" ON prec." + joinColumn + " = prec2." + joinColumn);
+        sql.append(" INNER JOIN targetedms.generalprecursor gp2 ");
+        sql.append("ON (gp2.Id = prec2.Id AND gp.Charge = gp2.Charge) ");
```
vagisha: I think we will need to add 'gp.mz = gp2.mz' in the join condition here.  The joinColumn ("IonFormula") in line 133 can be null.

vagisha (approved):
This was a lot of work!! Thank you so much for doing this.   I left a few comments. 
I will give this a try later this week, hopefully.

src/org/labkey/targetedms/query/ConflictResultsManager.java
labkey-jeckels: Changed this to concatenate the values, plus mz

resources/queries/targetedms/librarymoleculeprecursor/LibraryPrecursors.qview.xml
labkey-jeckels: Added MoleculeName as primary sort, but left the others as secondary ordering

src/org/labkey/targetedms/query/ConflictResultsManager.java
labkey-jeckels: Done

labkey-jeckels:
Thanks Vagisha. I addressed your concerns, which also led to to find some other problems, primarily related to how we match precursors with each other for small molecules.

labkey-ankurj (approved):

PR #280, opened 2020-11-10
Issue 41788: Exclude non-quantitative transitions from peak area ratio calculation

vagisha:
- Non-quantitative transition areas should be excluded when calculating peak area ratios stored in the targetedms schema.
- GeneralTransition.Quantitative column (and a few other recently added columns) are not getting populated when transition data is read from a Google Protocol Buffer
-  In Chromatogram charts intensities of points across the precursor peak are calculated by summing up the transition intensities at each point. Intensities of non-quantitative transitions should be excluded from this. If, however, none of the transitions for a precursor are quantitative then they should all be summed up


src/org/labkey/targetedms/query/TransitionManager.java
```diff
@@ -154,14 +154,16 @@ public static double getMaxTransitionIntensity(long generalMoleculeId, Transitio
     }
 
     @NotNull
-    public static Set<Integer> getTransitionChromatogramIndexes(long precursorChromInfoId)
+    public static Map<Integer, Boolean> getTransitionChromatogramIndexes(long precursorChromInfoId)
```
labkey-jeckels, line 157: Please add a comment on what the Map<> holds for values

src/org/labkey/targetedms/chart/ChromatogramDataset.java
```diff
@@ -525,25 +524,31 @@ private PeakInChart addPrecursorAsSeries(XYSeriesCollection dataset, Chromatogra
             double minTime = chromatogramRtRange.getMinRt();
             double maxTime = chromatogramRtRange.getMaxRt();
 
-            Set<Integer> transitionChromIndexes = TransitionManager.getTransitionChromatogramIndexes(pChromInfo.getId());
+            Map<Integer, Boolean> transitionChromIndexes = TransitionManager.getTransitionChromatogramIndexes(pChromInfo.getId());
+
+            // If none of the transitions are quantitative we will add them all.  Otherwise add only quantitative transitions.
+            boolean addAll = !transitionChromIndexes.entrySet().stream().anyMatch(k -> !Boolean.FALSE.equals(k.getValue()));
```
labkey-jeckels, line 530: I'm getting lost in the ! operators and it being equal to false. Is this because we may have null values in the map? Could this be:

```suggestion
            boolean addAll = transitionChromIndexes.values().stream().allMatch(v -> v == null || v.booleanValue()));
```

labkey-jeckels (changes requested):
A couple of suggestions. And yes, if this needs to be patched in 20.7 please go ahead and rename the branch and create a new PR against that branch.

src/org/labkey/targetedms/chart/ChromatogramDataset.java
vagisha: Yes, the map can have null values.   Quantitative transitions do not have the "quantitative" attribute in Skyline XML so those are all null in the database.  It is hopefully a little clearer in this PR for 20.7: https://github.com/LabKey/targetedms/pull/281

vagisha:
New PR against the 20.7 release: https://github.com/LabKey/targetedms/pull/281

PR #281, opened 2020-11-10
Issue 41788: Exclude non-quantitative transitions from peak area ratio calculation

vagisha:
- Non-quantitative transition areas should be excluded when calculating peak area ratios stored in the targetedms schema.
- GeneralTransition.Quantitative column (and a few other recently added columns) are not getting populated when transition data is read from a Google Protocol Buffer
-  In Chromatogram charts intensities of points across the precursor peak are calculated by summing up the transition intensities at each point. Intensities of non-quantitative transitions should be excluded from this. If, however, none of the transitions for a precursor are quantitative then they should all be summed up

Original PR against 20.11 release branch: https://github.com/LabKey/targetedms/pull/280


src/org/labkey/targetedms/parser/PeakAreaRatioCalculator.java
```diff
@@ -72,6 +72,11 @@ public void init(Map<SkylineDocImporter.SampleFileKey, SampleFile> skylineIdSamp
 
             for(Transition transition: precursor.getTransitionsList())
             {
+                // Issue 41788: Exclude non-quantitative transitions from peak area ratio calculations
+                if(!transition.isQuantitativeTransition())
```
nickshulman: When calculating total areas, we also skip over all of the MS2 transitions if the Transition Settings Full Scan Acquisition Method was "DDA".
In Skyline, this logic is taken care of in the method "bool TransitionDocNode.IsQuantitative(SrmSettings settings)".
The same sort of check should probably be happening here.

labkey-jeckels:
Looks OK from my side. I'll let you and Nick follow up as needed, and he can do the final PR approval.

src/org/labkey/targetedms/parser/PeakAreaRatioCalculator.java
vagisha: A few of questions about bool TransitionDocNode.IsQuantitative(SrmSettings settings)
- Does IsMs1() apply only to proteomic transitions? 
- ComplexFragmentIon.IsMS1() is looking at TransitionLosses.  We only save transition losses for proteomic transitions.  Should we be saving them for small molecule transitions as well? Should we move the variable from Transition to GeneralTransition?
- I don't think we are saving the children of ComplexFragmentIon so we will not be able to do this check: Children.Values.All(child => child.IsMs1

I will update the test with the document used in Skyline's Quantitative test.  If there is a better one let me know. 
Thanks!

nickshulman: For small molecules, a transition is MS1 if and only if its fragment ion is precursor.

For proteomic transitions, we also need to look at the neutral losses and complex fragment ion.

I made a trial implementation of this here:
https://github.com/nickshulman/targetedms/commit/c5639547b838239ad82f04d8f695bd9f403c20ef
(Haven't tested it, but it probably works)

vagisha: Thanks for the looksLikeIntactPrecursor() method!  Do you have a Skyline document I can test this with?

nickshulman (approved):

PR #283, opened 2020-11-12
Issue 41809: Multi-attribute Method configuration not saving for new Panorama folders

labkey-jeckels:
#### Rationale
The MAM folder type in Panorama isn't fully operational yet. The change to offer it during folder setup was never fully wired up

#### Changes
* Persist MAM folders as such
* Add automated test coverage for MAM reports


test/src/org/labkey/test/tests/targetedms/TargetedMSMAMTest.java
```diff
@@ -0,0 +1,533 @@
...
+    @LogMethod
+    protected void verifyAttributeGroupIdCalcs() throws IOException, CommandException
+    {
+        Connection cn = createDefaultConnection(false);
+        // Query for small molecule data
+        SelectRowsCommand smallMoleculeCommand = new SelectRowsCommand("targetedms", "generalmoleculechrominfo");
+        smallMoleculeCommand.setRequiredVersion(9.1);
+        smallMoleculeCommand.setColumns(Arrays.asList("MoleculeId/CustomIonName", "SampleFileId", "PeakCountRatio", "RetentionTime", "ModifiedAreaProportion", "SampleFileId/ReplicateId/RunId", "MoleculeId/AttributeGroupId", "PeptideId/AttributeGroupId"));
+        smallMoleculeCommand.addFilter("MoleculeId/AttributeGroupId", "C2X", Filter.Operator.EQUAL);
+        smallMoleculeCommand.addFilter("SampleFileId/SampleName", "13417_02_WAA283_3805_071514", Filter.Operator.EQUAL);
```
labkey-ankurj: does the filter get applied even when the column is not in columns list?

test/src/org/labkey/test/tests/targetedms/TargetedMSMAMTest.java
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    @LogMethod
+    protected void verifyModificationSearch()
+    {
+        // add modification search webpart and do an initial search by AminoAcid and DeltaMass
+        clickAndWait(Locator.linkContainingText("Panorama Dashboard"));
+        waitForElement(Locator.id("identifierInput"));
+        _ext4Helper.clickExt4Tab("Modification Search");
+        waitForElement(Locator.name("aminoAcids"));
```
labkey-ankurj: looks like strings "aminoAcids", "deltaMass", "unimodName", "customName"  are used in multiple places. can be pulled into variables?

test/src/org/labkey/test/tests/targetedms/TargetedMSMAMTest.java
```diff
@@ -0,0 +1,533 @@
...
+        setFormElement(Locator.name("aminoAcids"), "R");
+        setFormElement(Locator.name("deltaMass"), "10");
+        waitAndClickAndWait(Ext4Helper.Locators.ext4Button("Search"));
+        waitForText("Modification Search Results");
+        //waitForText("1 - 13 of 13");
+        assertTextPresentInThisOrder("Targeted MS Modification Search", "Targeted MS Peptides");
+        assertTextPresent("Amino acids:", "Delta mass:");
+        assertEquals(13, Locator.xpath("//td//a//span[contains(@title, 'R[+10.0]')]").findElements(getDriver()).size());
```
labkey-ankurj: good to make these strings 'R[+10.0]' and 'K[+8.0]' as variables as well.

labkey-ankurj (changes requested):

test/src/org/labkey/test/tests/targetedms/TargetedMSMAMTest.java
labkey-jeckels: I should have trimmed all of that code. It got copied over from the test class I copied, and is not used or needed in this new test. I just pushed a commit that deletes it. Sorry to make you review the full thing - the new test is really short!

labkey-ankurj (approved):

PR #286, opened 2020-11-18
Issue 36937: Add links to raw data in the targetedms replicates view

vagisha:
#### Rationale
Add links in the SampleFileTable grid to download raw data if it was uploaded.

#### Changes
- Added a "File" column to SampleFileTable that displays the file name extracted from SampleFile.FilePath
- Added a "Download" column with a link and download size
- Displaying links in the SampleFileTable only requires a matching row in the exp.data table.  Zip files (e.g. raw.zip for Waters data) are not validated for expected contents
- MsDataSource also has methods for validating zip files which will be used by the panoramapublic module.
- Tests added and updated; Dummy raw data added for tests


src/org/labkey/targetedms/MsDataSourceUtil.java
```diff
@@ -0,0 +1,1243 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import static org.hamcrest.CoreMatchers.hasItem;
+
+public class MsDataSourceUtil
```
labkey-jeckels: This a big class. I recommend creating a new package under org.labkey.targetedms (.raw, maybe), moving this into it, and pulling all of the nested classes into their own top-level classes.

src/org/labkey/targetedms/MsDataSourceUtil.java
```diff
@@ -0,0 +1,1243 @@
...
+                    }
+                    else
+                    {
+                        size = FileUtil.hasCloudScheme(dataPath) ? null : FileUtils.sizeOfDirectory(dataPath.toFile());
+                        return new RawDataInfo(expData, size, false);
+                    }
+                }
+                catch (IOException e)
+                {
+                    LOG.debug("Error getting size of " + dataPath, e);
```
labkey-jeckels: IOExceptions indicate there's something going wrong with the file system. debug() level messages will be effectively ignored via the logging configuration by default. I recommend throwing an UnexpectedException instead, wrapping the original IOException.

If there are specific cases where you want the error to be non-fatal, please log at least at the warn level so that it's more likely an admin will see it and correct the problem

src/org/labkey/targetedms/TargetedMSServiceImpl.java
```diff
@@ -228,4 +237,10 @@ public Integer importSkylineDocument(ViewBackgroundInfo info, Path skylinePath)
     {
         return TargetedMSManager.addRunToQueue(info, skylinePath);
     }
+
+    @Override
+    public List<? extends ISampleFile> hasData(List<? extends ISampleFile> sampleFiles, Container container)
```
labkey-jeckels: I'd expect a "has" method to return true/false. Make this getData() or getSampleFiles() instead?

src/org/labkey/targetedms/query/SampleFileTable.java
```diff
@@ -171,4 +205,80 @@ public QueryUpdateService getUpdateService()
...
+                .append(" WHEN ").append(filePathColSql).append(" IS NOT NULL AND ").append(dialect.getStringIndexOfFunction(new SQLFragment("'\\'"), filePathColSql)).append(" > 0")
+                .append(" THEN ").append(dialect.getSubstringFunction(
+                        filePathColSql,
+                        new SQLFragment(filePathLengthSql).append(" - ")
+                                .append(dialect.getStringIndexOfFunction(new SQLFragment("'\\'"), new SQLFragment(" REVERSE(").append(filePathColSql)).append(") + 2 ")),
+                        filePathLengthSql))
+                .append(" ELSE ").append(filePathColSql).append(" END");
+    }
+
+    private class DownloadLinkColumn extends DataColumn
```
labkey-jeckels, line 221: Should override isSortable() and isFilterable(), since they'll be operating on the underlying column value (a long)

labkey-jeckels (changes requested):

src/org/labkey/targetedms/MsDataSourceUtil.java
vagisha: Done. Moved to org.labkey.targetedms.datasource

src/org/labkey/targetedms/MsDataSourceUtil.java
vagisha: Done.

src/org/labkey/targetedms/query/SampleFileTable.java
vagisha: Done. Added isSortable() and isFilterable() overrides for the download column of the TargetedMS Runs table as well.

src/org/labkey/targetedms/TargetedMSServiceImpl.java
vagisha: I changed it to getSampleFilesWithData(). Hope that is OK.

labkey-jeckels (approved):

PR #296, opened 2020-12-10
Issue 41973: Populate additional information for rendering chromatograms

vagisha:
#### Changes
- Populate a BestMassErrorPPM column on PrecursorChromInfo so that we can label peaks when we are not saving TransitionChromInfos
- If we cannot determine the peak boundary (e.g. when TransitionChromInfos are not saved, and all the transitions for a precursor are non-quantitative) display the full chromatogram range
- Label precursor peak with the mass_error_ppm of the tallest transition rather than the averageMassErrorPpm set on the PrecursorChromInfo
- Non-quantitative transitions are determined with GeneralTransition.isQuantitative(TransitionSettings.FullScanSettings settings).  This will return false for MS2 transitions where the Full Scan Acquisition method is set to "DDA"
- Shorten y-axis (intensity) tick labels with a multiplier added to the axis label (e.g. Intensity 10^3).  This was not working.
- Allow up to two decimal places for RT axis to avoid repeated tick labels

### DIFF
--- a/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
+++ b/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
@@ -1683,56 +1683,43 @@ public class SkylineDocumentParser implements AutoCloseable
 
     private <T extends GeneralTransition> void updatePrecursorChromInfos(GeneralPrecursor<T> precursor)
     {
-        Map<String, List<TransitionChromInfo>> replicateToTciListMap = new HashMap<>();
+        TransitionSettings.FullScanSettings fullScanSettings = _transitionSettings == null ? null : _transitionSettings.getFullScanSettings();
+
+        // Key is the replicate name
+        // Value is a List where each member is a Pair of TransitionChromInfo and a Boolean value indicating if the transition is quantitative
+        Map<String, List<Pair<TransitionChromInfo, Boolean>>> replicateToTciListMap = new HashMap<>();
+
         for (GeneralTransition transition : precursor.getTransitionsList())
         {
             List<TransitionChromInfo> tciList = transition.getChromInfoList();
+            Boolean quantitative = transition.isQuantitative(fullScanSettings);
             for(TransitionChromInfo tci: tciList)
             {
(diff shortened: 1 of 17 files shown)


src/org/labkey/targetedms/query/PrecursorManager.java
```diff
@@ -605,6 +601,28 @@ public static double getMaxPrecursorIntensity(long generalMoleculeId)
         return new SqlSelector(TargetedMSManager.getSchema(), sql).getObject(Double.class);
     }
 
+    /**
+     * Returns the maximum value of MaxHeight for the Precursors of a peptide (generalMolecule) over all the replicates.
+     * MaxHeight of a PrecursorChromInfo is the height of the tallest transition peak for the precursor in a replicate.
+     * This can be used for getting the height of the tallest transition peak for a peptide over all the replicates
+     * when we are synchronizing the intensity axis for transition peak chromatograms.
+     * @param generalMoleculeId
```
labkey-jeckels: These two lines can be trimmed - they're not adding anything

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -1676,9 +1676,54 @@ else if (XmlUtil.isStartElement(reader, evtType, NOTE))
 
         _precursorCount++;
 
+        updatePrecursorChromInfos(moleculePrecursor);
+
         return moleculePrecursor;
     }
 
+    private <T extends GeneralTransition> void updatePrecursorChromInfos(GeneralPrecursor<T> precursor)
```
labkey-jeckels: Does this need to return the transitions? It looks like they're not used.

The method name is kind of ambiguous. How about updateChromatogramSettings() or something like that, and/or a comment about what it's doing?

labkey-jeckels (approved):
A couple suggestions related to comments and method naming but change itself looks good!

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
vagisha: The method does not return anything. But the type parameter is confusing and can be replaced with a wildcard.

PR #303, opened 2020-12-18
Attempt to fix intermittent test failure on TargetedMSRun leak

labkey-jeckels:
#### Rationale
We've seen at least one downstream test fail with a reported leak of a TargetedMSRun from MsDataSourceUtil.TestCase

https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_BvtBSqlserver/1200925?buildTab=overview

#### Changes
* Null out the reference, clean up exception handling a bit


labkey-tchad (approved):
This was actually always leaking. There just happened to be a bug in the test harness that was causing it to usually be missed.

PR #319, opened 2021-01-20
Item 8311: Correlate QC and experimental -Instrument Summary Page

labkey-ankurj:
#### Rationale
This work links the QC folders that have data from instruments.

#### Changes
* New InstrumentSummary web part
* New showInstrument action

### DIFF
--- a/resources/queries/targetedms/InstrumentSummaryByFolder.sql
+++ b/resources/queries/targetedms/InstrumentSummaryByFolder.sql
@@ -6,8 +6,6 @@ SELECT
        ReplicateId.RunId.Container,
        InstrumentSerialNumber
 FROM targetedms.SampleFile
-WHERE
-      InstrumentSerialNumber IS NOT NULL
 GROUP BY
          ReplicateId.RunId.Container,
          InstrumentSerialNumber
--- a/resources/queries/targetedms/InstrumentSummaryByFolder.sql
+++ b/resources/queries/targetedms/InstrumentSummaryByFolder.sql
@@ -1,0 +1,13 @@
+SELECT
+       COUNT(ReplicateId.RunId) AS SkylineDocumentCount,
+       COUNT(DISTINCT ReplicateId) AS ReplicateCount,
+       MIN(AcquiredTime) AS FirstAcquisition,
+       MAX(AcquiredTime) AS LastAcquisition,
+       ReplicateId.RunId.Container,
+       InstrumentSerialNumber
+FROM targetedms.SampleFile
+WHERE
+      InstrumentSerialNumber IS NOT NULL
+GROUP BY
+         ReplicateId.RunId.Container,
+         InstrumentSerialNumber
--- a/resources/queries/targetedms/QCInstrumentSummary.sql
+++ b/resources/queries/targetedms/QCInstrumentSummary.sql
@@ -3,7 +3,8 @@ SELECT
     sf.InstrumentSerialNumber AS SerialNumber,
     MIN(sf.AcquiredTime) AS StartDate,
     MAX(sf.AcquiredTime) AS EndDate,
-    COUNT(DISTINCT sf.ReplicateId) AS NoOfReplicates
+    COUNT(DISTINCT sf.ReplicateId) AS NoOfReplicates,
+    null AS QCFolders
 FROM targetedms.SampleFile sf
 WHERE
     sf.InstrumentSerialNumber IS NOT NULL
--- a/resources/queries/targetedms/QCInstrumentSummary.sql
+++ b/resources/queries/targetedms/QCInstrumentSummary.sql
@@ -1,0 +1,12 @@
+SELECT
+    sf.InstrumentId.model AS InstrumentName,
+    sf.InstrumentSerialNumber AS SerialNumber,
+    MIN(sf.AcquiredTime) AS StartDate,
+    MAX(sf.AcquiredTime) AS EndDate,
+    COUNT(DISTINCT sf.ReplicateId) AS NoOfReplicates
+FROM targetedms.SampleFile sf
+WHERE
+    sf.InstrumentSerialNumber IS NOT NULL
+GROUP BY
+         sf.InstrumentSerialNumber,
+         sf.InstrumentId.model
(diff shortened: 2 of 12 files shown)


resources/queries/targetedms/InstrumentSummaryByFolder.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+       COUNT(ReplicateId.RunId) AS SkylineDocumentCount,
+       COUNT(DISTINCT ReplicateId) AS ReplicateCount,
+       MIN(AcquiredTime) AS FirstAcquisition,
+       MAX(AcquiredTime) AS LastAcquisition,
+       ReplicateId.RunId.Container,
+       InstrumentSerialNumber
+FROM targetedms.SampleFile
+WHERE
+      InstrumentSerialNumber IS NOT NULL
```
labkey-jeckels: I'd like to still show files without a serial number in the summary even if we can't offer QC folder links.

resources/queries/targetedms/QCInstrumentSummary.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+    sf.InstrumentId.model AS InstrumentName,
+    sf.InstrumentSerialNumber AS SerialNumber,
+    MIN(sf.AcquiredTime) AS StartDate,
+    MAX(sf.AcquiredTime) AS EndDate,
+    COUNT(DISTINCT sf.ReplicateId) AS NoOfReplicates,
+    null AS QCFolders
```
labkey-jeckels: Instead of null, have this use sf.InstrumentSerialNumber too.

src/org/labkey/targetedms/query/QCFolderDisplayColumnFactory.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public DisplayColumn createRenderer(ColumnInfo colInfo)
+    {
+        return new DataColumn(colInfo)
+        {
+            @Override
+            public void renderGridCellContents(RenderContext ctx, Writer out) throws IOException
+            {
+                final User user = ctx.getViewContext().getUser();
+                String serialNumber = String.valueOf(ctx.get("SerialNumber"));
```
labkey-jeckels: We discussed yesterday that this value won't be available if someone customizes the view and removes the column.

See my suggestion above of aliasing the serial number as the value of this column. Then you can use getBoundColumn().getValue(ctx) to get the value in this code without needing to worry about the addQueryFieldKeys() stuff

src/org/labkey/targetedms/query/QCFolderDisplayColumnFactory.java
```diff
@@ -0,0 +1,65 @@
...
+                var instrumentRunIds = TargetedMSManager.getRunIdsByInstrument(serialNumber);
+                List<Container> qcContainers = new ArrayList<>();
```
labkey-jeckels: Use a Set to eliminate duplicates

src/org/labkey/targetedms/query/QCFolderDisplayColumnFactory.java
```diff
@@ -0,0 +1,65 @@
...
+                        {
+                            qcContainers.add(run.getContainer());
+                        }
+                    }
+                });
+                StringBuilder sb = new StringBuilder();
+                qcContainers.forEach(qcContainer -> {
+                    var url = qcContainer.getStartURL(user);
+                    sb.append("<div><a href=\"")
+                            .append(url)
```
labkey-jeckels: Go ahead and filter this too, just to be paranoid

src/org/labkey/targetedms/TargetedMSModule.java
```diff
@@ -411,6 +413,15 @@ public boolean isExactMatch()
                     v.setFrame(WebPartView.FrameType.PORTAL);
                     return v;
                 }
+            },
+
+            new BaseWebPartFactory(TARGETED_MS_RUNS_WEBPART_NAME)
```
labkey-jeckels: This doesn't seem right - presumably this should be TARGETED_MS_INSTRUMENT_SUMMARY

labkey-jeckels (changes requested):

src/org/labkey/targetedms/query/QCFolderDisplayColumnFactory.java
labkey-ankurj: OK

src/org/labkey/targetedms/query/QCFolderDisplayColumnFactory.java
labkey-ankurj: OK

src/org/labkey/targetedms/query/QCFolderDisplayColumnFactory.java
labkey-ankurj: Sounds good

src/org/labkey/targetedms/TargetedMSModule.java
labkey-ankurj: oops, corrected

resources/queries/targetedms/QCInstrumentSummary.sql
labkey-ankurj: done

resources/queries/targetedms/InstrumentSummaryByFolder.sql
labkey-ankurj: OK

src/org/labkey/targetedms/query/QCFolderDisplayColumnFactory.java
```diff
@@ -35,9 +35,9 @@ public DisplayColumn createRenderer(ColumnInfo colInfo)
             public void renderGridCellContents(RenderContext ctx, Writer out) throws IOException
             {
                 final User user = ctx.getViewContext().getUser();
-                String serialNumber = String.valueOf(ctx.get("SerialNumber"));
+                String serialNumber = String.valueOf(getBoundColumn().getValue(ctx));
                 var instrumentRunIds = TargetedMSManager.getRunIdsByInstrument(serialNumber);
-                List<Container> qcContainers = new ArrayList<>();
+                Set<Container> qcContainers = new HashSet<>();
```
labkey-jeckels: Even better, let's use a TreeSet which will sort them in a consistent way.

labkey-jeckels (approved):
One more simple suggestion but looks good. No further review needed.

PR #321, opened 2021-01-25
Issue 42317: NPE in download column of replicates grid when instrument info is not available for a sample file

vagisha:
#### Rationale
Instrument info may not be available for replicates imported with very old versions of Skyline.

### DIFF
--- a/src/org/labkey/targetedms/datasource/MsDataSourceUtil.java
+++ b/src/org/labkey/targetedms/datasource/MsDataSourceUtil.java
@@ -203,9 +203,9 @@ public class MsDataSourceUtil
         {
             // We can get more than one source type by filename extension lookup. e.g. .raw extension is used both by Thermo and Waters
             // Try to resolve by looking up the instrument on which the data was acquired.
-            Instrument instrument = InstrumentManager.getInstrument(sampleFile.getInstrumentId());
-            if(instrument != null)
+            if(sampleFile.getInstrumentId() != null)
             {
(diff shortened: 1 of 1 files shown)


src/org/labkey/targetedms/datasource/MsDataSourceUtil.java
```diff
@@ -203,9 +203,9 @@ else if(sourceTypes.size() > 1)
         {
             // We can get more than one source type by filename extension lookup. e.g. .raw extension is used both by Thermo and Waters
             // Try to resolve by looking up the instrument on which the data was acquired.
-            Instrument instrument = InstrumentManager.getInstrument(sampleFile.getInstrumentId());
-            if(instrument != null)
+            if(sampleFile.getInstrumentId() != null)
             {
+                Instrument instrument = InstrumentManager.getInstrument(sampleFile.getInstrumentId());
```
labkey-jeckels, line 208: This is probably fine because the foreign key should ensure there's a matching row in the Instrument table, but just wanted to confirm you intentionally removed the null check on instrument.

labkey-jeckels (approved):

src/org/labkey/targetedms/datasource/MsDataSourceUtil.java
vagisha: Yes, if there is an instrument ID a matching row should exist in the Instrument table.

PR #325, opened 2021-01-29
Automation test for Chromatogram optimization library.

labkey-sweta:
#### Rationale
Automation test for Chromatogram optimization library. Reading SQLite file.

#### Related Pull Requests
*https://github.com/LabKey/testAutomation/pull/614

#### Changes
1. test/sampledata/TargetedMS/SmallMoleculeLibrary3.sky.zip
2. test/src/org/labkey/test/util/targetedms/ConnectionSource.java
3. test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java


test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.labkey.test.util.targetedms.ConnectionSource;
+
+import java.io.File;
+import java.net.URL;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+@Category({DailyB.class, MS2.class})
```
labkey-jeckels: Panorama/TargetedMS is separate from the MS2 module. I know other TargetedMS tests are tagged as MS2. I don't know what the suite is intended to represent, but it should be renamed or slimmed down.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,133 @@
...
+    @Test
+    public void testUpload() throws Exception
+    {
+        goToProjectHome();
+        importData(SKY_FILE);
+        clickAndWait(Locator.linkContainingText("Panorama Dashboard"));
+        File downloadedClibFile = doAndWaitForDownload(() -> clickButton("Download", 0));
+
+        goToProjectHome();
+        String queryURL = "query-executeQuery.view?schemaName=targetedms&query.queryName=TransitionOptimization";
```
labkey-jeckels: Is there a reason this is navigating by direct URL? I was under the impression that we shouldn't be hard-coding URLs in tests except in very special circumstances

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,133 @@
...
+            DatabaseMetaData md = conn.getMetaData();
+            ResultSet rs = md.getTables(null, null, tableName, null);
+            if (rs.next())
+                return true; //Table exists
+            else
+                return false;
+        }
+        catch (SQLException e)
+        {
+            e.printStackTrace();
```
labkey-jeckels: This exception shouldn't be caught and swallowed like this. I recommend just letting it bubble up through the call stack

test/src/org/labkey/test/util/targetedms/ConnectionSource.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public ConnectionSource(String libraryFilePath)
+    {
+        // Create an ObjectPool that serves as the pool of connections.
+        _connectionPool = new GenericObjectPool();
+        _connectionPool.setMaxActive(5);
+
+        // Create a ConnectionFactory that the pool will use to create Connections.
+        ConnectionFactory connectionFactory = new DriverManagerConnectionFactory("jdbc:sqlite:/" + libraryFilePath, null);
```
labkey-jeckels: We shouldn't need to use a connection pool here. A simple Connection should be sufficient, and allow us to avoid the dependency on DBCP.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+        return false;
+    }
+
+    private boolean columnExists(File clibFile, String tableName, String columnName)
+    {
+        ConnectionSource cs = new ConnectionSource(clibFile.getAbsolutePath());
+        try
+        {
+            Connection conn = cs.getConnection();
```
labkey-jeckels: Please use a try-with-resources syntax to be sure that the connections are closed.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,133 @@
...
+            DatabaseMetaData md = conn.getMetaData();
+            ResultSet rs = md.getColumns(null, null, tableName, columnName);
+            if (rs.next())
```
labkey-jeckels: Same with the ResultSets - use a try-with-resources

labkey-jeckels (changes requested):

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: Done.!

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: Done.!

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: Done

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: Removed..!

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: Yes, i was navigating directly because there is was a case mismatch. Corrected the XML file.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,126 @@
...
+            DatabaseMetaData md = conn.getMetaData();
+            ResultSet rs = md.getTables(null, null, tableName, null);
+            if (rs.next())
+                return true; //Table exists
+            else
+                return false;
+        }
+        catch (SQLException e)
+        {
+            throw new RuntimeException(e);
```
labkey-jeckels: This is an improvement over just catching the exception and printing the stack trace, but you can declare that these methods throw SQLException and not have to catch and rethrow them. The method that's calling them is already throwing Exception, so this will reduce the code footprint.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+import java.io.File;
+import java.net.URL;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+@Category({DailyB.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 6)
```
labkey-jeckels: Is this timeout meaningful? Can we just leave it with the default value?

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,126 @@
...
+        importData(SKY_FILE);
+        clickAndWait(Locator.linkContainingText("Panorama Dashboard"));
+        File downloadedClibFile = doAndWaitForDownload(() -> clickButton("Download", 0));
+
+        goToProjectHome();
+        goToSchemaBrowser();
+        DataRegionTable table = viewQueryData("targetedms", "TransitionOptimization");
+        table.rowSelector().showAll();
+
+        checker().verifyEquals("Invalid number of rows in transition optimization", table.getDataRowCount(),
```
labkey-jeckels: This ordering seems odd. First we check the number of rows in the table, and then we separately check that the table exists?

Checking a few other tables would be good, as this test coverage is new for the generated SQLite files so there's some catchup to do.

labkey-jeckels (changes requested):
Thanks for the improvements. A few more suggestions but that should do it.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: Done.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: The test took 2.14 sec to run on teamcity, so changed the timeout to 3.
https://teamcity.labkey.org/viewLog.html?buildId=1245005&buildTypeId=LabkeyTrunk_ModuleSuites_TargettedMSPostgres&tab=testsInfo

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: I am verifying the tables mentioned in the spec, Do u know any other table to be verified?
@labkey-jeckels @labkey-ankurj

labkey-jeckels: See here for the schema:

https://docs.google.com/document/d/1U1Oiy6LaWRqd5EaJ0h-WHFTLrWg360yUdcXAj4XumzY/edit#heading=h.sxp2n0h84m9u

Or you can use a SQLite file browser (including IntelliJ) to crack open the .clib and see what it contains. This is for PyCharm instead of IntelliJ but the same drag/drop works great in both.

https://www.jetbrains.com/pycharm/guide/tips/create-sqlite-connection/

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,105 @@
...
+        ResultSet rs = md.getTables(null, null, tableName, null);
+        if (rs.next())
+            return true; //Table exists
+        else
+            return false;
+    }
+
+    private boolean columnExists(File clibFile, String tableName, String columnName) throws SQLException
+    {
+        try (Connection conn = ConnectionSource.getConnection(clibFile.getAbsolutePath()))
```
labkey-jeckels, line 102: Your changes to the exception handling removed the code that was making sure these connections and result sets get closed properly. I just pushed a correction for the columnExists() method. Please apply the same pattern to the others.

labkey-jeckels (changes requested):
See comment about closing the resources and my link to schema info.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: Done.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: added coverage based on the email reply.

labkey-jeckels: @labkey-sweta Please push your changes. They're not showing in GitHub yet.

labkey-sweta: My push was rejected because the my branch was behind, should be there now.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private int getServerTableRowCount(String tableName)
+    {
+        goToSchemaBrowser();
+        DataRegionTable table = viewQueryData("targetedms", tableName);
+        table.rowSelector().showAll();
+        try
+        {
+            table.goToView("LibraryMolecules");
```
labkey-jeckels: Seems like it would be cleaner to pass an optional viewName parameter to this method. If it's not null, the test would switch to that view.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private List<String> tableExists(List<String> tables, File clibFile) throws SQLException
+    {
+        try (Connection conn = ConnectionSource.getConnection(clibFile.getAbsolutePath()))
+        {
+            DatabaseMetaData md = conn.getMetaData();
+            for (Iterator<String> i = tables.iterator(); i.hasNext(); )
+            {
+                String tableName = i.next();
+                ResultSet rs = md.getTables(null, null, tableName, null);
```
labkey-jeckels: These ResultSets should be closed as well via a try-with-resource syntax. Your push overwrote my example that I pushed for columnExists(). See https://github.com/LabKey/targetedms/pull/325/commits/53c56b78ea54262089151cc9c4b732b748e8c828

labkey-jeckels (changes requested):
One more round, sorry.

test/src/org/labkey/test/tests/targetedms/TargetedMSChromatogramOptimizationTest.java
labkey-sweta: Yup, sorry for that.

labkey-jeckels (approved):

PR #326, opened 2021-01-31
Changes necessary to support the complex crosslinks feature that will…

nickshulman:
… be available in Skyline 21.1

#### Rationale
Skyline 21.1 is going to support more complicated crosslinks.
As a result, the way that Skyline refers to the crosslinked peptides is different.
In Skyline 20.2, the peptides were referred to by the amino acid index and modification name where they attached to the main peptide (e.g. "3:disulfide"), but now it's just a flat list of peptide sequence which get referred to with just a number (where "0" refers to the main peptide).
Similarly, fragment ions are just a list of the ion types they are (e.g. "p-y7")

#### Related Pull Requests
https://github.com/ProteoWizard/pwiz/pull/1428

#### Changes
In Skyline 20.2, the peptides were referred to by the amino acid index and modification name where they attached to the main peptide (e.g. "3:disulfide"), but now it's just a flat list of peptide sequence which get referred to with just a number (where "0" refers to the main peptide).
Similarly, fragment ions are just a list of the ion types they are (e.g. "p-y7")

Also: there was a bug in "transitionProtoToTransition" where it was using "transition.getFragmentType()", which does not actually get set until later on by the caller.
As a result of this, the complex fragment ion names always started with "-" instead of the actual type of the first part of the ion.

The two tests that were there: "testCrosslinkingDocument" and "testCrosslinkingCompactDocument" are testing what ought to be identical documents, but because of this bug they were not identical. I changed the body of the tests so that they now assert the same things.


labkey-jeckels (approved):
These changes look good to me.

My one question is whether we'd want to migrate values that have already been imported, and transform future imports to the new format. It doesn't look too bad to maintain support for the legacy format so probably not worth it.

labkey-jeckels:
@nickshulman is this getting ready to merge? We'll be branching for our 21.3 release at the end of the month.

nickshulman:
Closing this pull request because it's from the nickshulman fork. Opened a different pull request which is not in a fork.

PR #331, opened 2021-02-05
Item 8348 : Link QC and experimental data - display experiment run range region on QC plots

labkey-ankurj:
#### Rationale
This PR is continuation of https://github.com/LabKey/targetedms/pull/319 and this work highlights the experiment run region when the linked QC folder is clicked from Panorama experimental folder.

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/319


src/org/labkey/targetedms/view/InstrumentSummaryWebPart.java
```diff
@@ -11,7 +12,14 @@
     public InstrumentSummaryWebPart (ViewContext viewContext)
     {
         super(new TargetedMSSchema(viewContext.getUser(), viewContext.getContainer()));
-        setSettings(new QuerySettings(getViewContext(), "InstrumentSummary", "QCInstrumentSummary"));
+        QuerySettings instrumentSummaryQS = new QuerySettings(getViewContext(), "InstrumentSummary", "QCInstrumentSummary");
+        var propertyValues = getBindPropertyValues();
+        if (null != propertyValues)
+        {
+            var runId = propertyValues.getPropertyValue("id").getValue().toString();
```
labkey-jeckels: Can the value of "id" ever be null?

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -543,6 +555,35 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
 
+    getExperimentRunDateRangeToolbar : function()
+    {
+        if (!this.experimentRunDateRangeToolbar)
+        {
+            var hidden = !this.showExpRunRange;
+            var returnUrl = LABKEY.ActionURL.getReturnUrl();
+            var htmlStr = this.showExpRunRange
+                    ? "<a href=" + returnUrl +  ">"
+                    + this.expRunDetails.fileName + " : "
```
labkey-jeckels: Need to HTML encode values

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1529,20 +1570,65 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
+                        .enter().append("rect").attr("class", "expRange")
+                        .attr('x', xAcc).attr('y', yRange[1])
+                        .attr('width', widthAcc).attr('height', yRange[0] - yRange[1])
+                        .attr('stroke', '#557098').attr('stroke-opacity', 0.1)
+                        .attr('fill', '#557098').attr('fill-opacity', 0.1);
+
+                // TODO: look into setting background color of title tooltip
+                expRange.append("title").text(function (d) {
+                    return "Skyline File: " + me.expRunDetails.fileName
+                            + ", \nSerial No: " + me.expRunDetails.serialNumber
```
labkey-jeckels: Need to HTML encode values

src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -3698,6 +3695,14 @@ protected QueryView createQueryView(RunDetailsForm form, BindException errors, b
             TargetedMSSchema schema = new TargetedMSSchema(getUser(), getContainer());
             return schema.createView(getViewContext(), settings, errors);
         }
+
+        @Override
+        protected ModelAndView getHtmlView(RunDetailsForm form, BindException errors) throws Exception
+        {
+            VBox vBox = (VBox) super.getHtmlView(form, errors);
+            vBox.addView(new InstrumentSummaryWebPart(form.getViewContext()), 1);
```
labkey-jeckels: We need to make this go through createQueryView() to create the instrument summary grid. Otherwise, when you export the grid it won't be filtered the same.

This line should call createQueryView() and pass in a name for the data region.

Then make the createQueryView() method check for the name and hand out the appropriate view. InstrumentSummaryWebPart is already a QueryView so it should be easy. If you pass in RunDetailsForm instead of the ViewContext, you'll have easy access to an already-parsed run id.

labkey-jeckels (changes requested):

src/org/labkey/targetedms/view/InstrumentSummaryWebPart.java
labkey-ankurj: no, I checked during dev, when the Id is null or not present, the action redirects to showList

src/org/labkey/targetedms/TargetedMSController.java
labkey-ankurj: converted to createQueryView, can not pass RunDetailsForm in the existing constructor as the InstrumentSummaryWebPart is being constructed as WebPartFactory.

webapp/TargetedMS/js/QCTrendPlotPanel.js
labkey-ankurj: done

webapp/TargetedMS/js/QCTrendPlotPanel.js
labkey-ankurj: done

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -563,9 +563,9 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
             var returnUrl = LABKEY.ActionURL.getReturnUrl();
             var htmlStr = this.showExpRunRange
                     ? "<a href=" + returnUrl +  ">"
```
labkey-jeckels: Don't forget to encode this one too.

src/org/labkey/targetedms/view/InstrumentSummaryWebPart.java
labkey-jeckels: That's true if you're going to the replicate page for a Skyline document, but I think this will break the use case of the web part on a portal page. I tested and getPropertyValue("id") returns null there if you don't have a "id" property on the URL

labkey-jeckels (changes requested):

src/org/labkey/targetedms/query/QCFolderDisplayColumnFactory.java
```diff
@@ -51,10 +52,15 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                 StringBuilder sb = new StringBuilder();
                 qcContainers.forEach(qcContainer -> {
                     var url = qcContainer.getStartURL(user);
+                    if (!"null".equals(currentRunId))
```
labkey-jeckels: Please don't use "null" as an actual string value. It's much cleaner to use a null value itself. It'll mean you need to change how you get the value on line 39 too

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -184,10 +184,21 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
 
-        // initialize the form panel toolbars and display the plot
-        this.add(this.initPlotFormToolbars());
+        this.getExpRunRangeDetails();
+    },
 
-        this.displayTrendPlot();
+    getExpRunRangeDetails: function() {
+        var urlParams = LABKEY.ActionURL.getParameters();
+        this.showExpRunRange = urlParams['RunId'] !== undefined && urlParams['RunId'] !== "null" && urlParams['RunId'] !== "";
```
labkey-jeckels: Can we avoid the "null" string here too? What are you guarding against? If you're concerned about bogus values on the URL, I assume non-numerics would trip up the code too.

labkey-jeckels (changes requested):

labkey-jeckels (approved):
Great!

PR #332, opened 2021-02-09
Small molecule chromatogram library column renaming

labkey-jeckels:
#### Rationale
We can freely improve naming in the small molecule chromatogram schema for CLIB files since they're not in use in the wild yet. Also, we can consolidate the DriftTimeXXX columns on PrecursorChromInfo and TransitionChromInfo into the newer IonMobilityXXX columns

#### Changes
* Rename from CustomIonName->MoleculeName and IonFormula->ChemicalFormula
* Extract Adduct part of MoleculePrecursor formula
* Migrate any existing Panorama DriftTimeXXX values into IonMobilityXXX columns and drop the old ones

### DIFF
--- a/src/org/labkey/targetedms/chromlib/Constants.java
+++ b/src/org/labkey/targetedms/chromlib/Constants.java
@@ -178,6 +178,7 @@ class Constants
         Fwhm("DOUBLE"),
         MassErrorPPM("DOUBLE"),
         ChromatogramIndex("INTEGER"),
+        FragmentName("VARCHAR(100)"),
 
         IrtLibraryId("INTEGER NOT NULL", Table.IrtLibrary, Id),
         PeptideModSeq("TEXT NOT NULL"),
@@ -186,6 +187,7 @@ class Constants
         TimeSource("INT"),
 
         TransitionId("INTEGER NOT NULL", Table.Transition, Id),
+        MoleculeTransitionId("INTEGER NOT NULL", Table.MoleculeTransition, Id),
         OptimizationType("TEXT NOT NULL"),
         OptimizationValue("DOUBLE NOT NULL"),
         OptimizationStep("INTEGER"),
@@ -824,8 +826,9 @@ class Constants
     {
         Id(Column.Id),
         MoleculePrecursorId(Column.MoleculePrecursorId),
-        MoleculeName(Column.MoleculeName),
+        FragmentName(Column.MoleculeName),
         ChemicalFormula(Column.ChemicalFormula),
+        Adduct(Column.Adduct),
         Mz(Column.Mz, "DOUBLE"),
         Charge(Column.Charge, "INTEGER"),
         FragmentType(Column.FragmentType),
@@ -925,7 +928,7 @@ class Constants
     public enum MoleculeTransitionOptimizationColumn implements ColumnDef
     {
         Id(Column.Id),
-        TransitionId(Column.TransitionId),
+        MoleculeTransitionId(Column.MoleculeTransitionId),
         OptimizationType(Column.OptimizationType),
         OptimizationValue(Column.OptimizationValue);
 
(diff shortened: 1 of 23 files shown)


src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -2195,13 +2195,26 @@ private PrecursorChromInfo readPrecursorChromInfo(XMLStreamReader reader) throws
...
+
+        // Support old Skyline documents that used "drift time" instead of "ion mobility"
+        Double driftTimeMs1 = XmlUtil.readDoubleAttribute(reader, "drift_time_ms1");
+        Double driftTimeWindow = XmlUtil.readDoubleAttribute(reader, "drift_time_window");
+        Double driftTimeFragment = XmlUtil.readDoubleAttribute(reader, "drift_time_fragment");
+
+        if (driftTimeMs1 != null || driftTimeWindow != null || driftTimeFragment != null)
+        {
+            chromInfo.setIonMobilityMs1(driftTimeMs1);
+            chromInfo.setIonMobilityMs1(driftTimeFragment);
```
labkey-ankurj:
```suggestion
            chromInfo.setIonMobilityFragment(driftTimeFragment);
```

labkey-ankurj (changes requested):

src/org/labkey/targetedms/chromlib/LibMolecule.java
```diff
@@ -3,30 +3,30 @@
 public class LibMolecule extends AbstractLibMolecule<LibMoleculePrecursor>
 {
     private Integer _moleculeListId;
-    private String _ionFormula;
-    private String _customIonName;
+    private String _chemicalFormula;
+    private String _moleculeName;
```
labkey-ankurj, line 7: does this need to be renamed as fragmentName?

src/org/labkey/targetedms/chromlib/Constants.java
```diff
@@ -121,8 +121,8 @@ private Constants() {}
         CalcNeutralMass("DOUBLE NOT NULL"),
         NumMissedCleavages("INTEGER NOT NULL"),
 
-        IonFormula("VARCHAR(100)"),
-        CustomIonName("VARCHAR(100)"),
+        ChemicalFormula("VARCHAR(100)"),
+        MoleculeName("VARCHAR(100)"),
```
labkey-ankurj, line 125: I suggest to remove this and use FragmentName defined at L181 in MoleculeTransitionColumn. FragmentName column

src/org/labkey/targetedms/chromlib/Constants.java
```diff
@@ -710,8 +707,8 @@ public String definition()
     {
         Id(Column.Id),
         MoleculeListId(Column.MoleculeListId),
-        IonFormula(Column.IonFormula),
-        CustomIonName(Column.CustomIonName),
+        IonFormula(Column.ChemicalFormula),
+        CustomIonName(Column.MoleculeName),
```
labkey-ankurj: may be use Column.FragmentName

labkey-ankurj (approved):
couple of additional renaming comments

src/org/labkey/targetedms/chromlib/LibMolecule.java
labkey-jeckels: Brian requested that we keep the molecule one as MoleculeName and call the MoleculeTransition one as FragmentName, which I think is the case as currently coded.

labkey-ankurj:
ah yes, got it. thanks

src/org/labkey/targetedms/chromlib/Constants.java
labkey-jeckels: This pattern's a little clunky, but the two names should agree with each other. I found a couple of cases that were out of sync, including this one.

PR #337, opened 2021-02-16
Item 8417: Show reference guideset to inform the selected date range

labkey-ankurj:
#### Rationale
If the date range(s) for the Guide Set(s) are too far apart from the selected dates, data from the dates in between will be truncated so that the Guide Set can be displayed together with the QC data from the custom dates.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2002

#### Changes
* New checkbox option to see reference guide set

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -1307,7 +1307,7 @@ public class TargetedMSController extends SpringActionController
             GuideSet gs = null;
             for (GuideSet guideSet : guideSets)
             {
-                if (guideSet.getTrainingEnd().before(startDate))
+                if (!guideSet.isDefault() && guideSet.getTrainingEnd().before(startDate))
                 {
                     if (null == gs)
                     {
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -1270,7 +1270,11 @@ public class TargetedMSController extends SpringActionController
             GuideSet closestOutOfRangeGuideSet = null;
             if (form.isShowOutOfRangeGS() && !guideSets.isEmpty())
             {
-                closestOutOfRangeGuideSet = TargetedMSManager.getClosestPastGuideSet(guideSets, form.getStartDate());
+                closestOutOfRangeGuideSet = getClosestPastGuideSet(guideSets, form.getStartDate());
+                if (null != closestOutOfRangeGuideSet)
+                {
+                    form.setStartDate(new java.sql.Date(closestOutOfRangeGuideSet.getTrainingStart().getTime()));
+                }
             }
 
             List<QCMetricConfiguration> qcMetricConfigurations = TargetedMSManager.get()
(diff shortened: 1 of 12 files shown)


src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -1137,6 +1137,7 @@ public Object execute(QCMetricOutliersForm form, BindException errors)
         private Date _endDate;
         private List<OutlierGenerator.AnnotationGroup> _selectedAnnotations;
         private boolean _showExcluded;
+        private boolean _showOutOfRangeGS;
```
labkey-jeckels: Let's rename to _showReferencedGS or similar to match the new text in the UI

webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -175,6 +178,27 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
...
+
+            if (this.filterQCPoints && !this.groupedX) {
+                for (var j = 0; j < plotDataRow.data.length; j++) {
+                    var plotData = plotDataRow.data[j];
+
+                    // for truncating out of range guideset data find last index of plotData starting from this.startDate
+                    if (plotData.AcquiredTime >= this.startDate) {
+                        if (!this.outOfRangeGSLastIndex) {
+                            // skip 5 points
+                            this.outOfRangeGSLastIndex = j - 5;
```
labkey-jeckels: Will we always have five to skip? Could we end up walking off the end of the array?

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1065,14 +1033,35 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
         return this.showExcludedPointsCheckbox;
     },
 
-    getGuideSetCreateButton : function()
-    {
+    getShowOutOfRangeGSCheckbox : function() {
+        if (!this.showOutOfRangeGSCheckbox) {
```
labkey-jeckels: These variables and methods could be renamed too

labkey-jeckels (changes requested):
I've got other local changes so I didn't pull this to test the changes, but I think there are two things in the spec that are missing:

1. Some visual indicator that data points/dates were dropped. A vertical line or something like that would be fine. I imagine it would be similar to the white overlay that you were doing.
2. New metrics to help gauge adoption based on the success criteria in the spec.

labkey-ankurj:
@labkey-jeckels does this separator look good? 
Screen Shot 2021-02-18 at 12 30 07 PM

labkey-jeckels:
@labkey-ankurj yes, thanks

labkey-jeckels (commented):
@labkey-ankurj did you see my question about subtracting five indices? Will that always be safe?

webapp/TargetedMS/js/QCPlotHelperBase.js
labkey-ankurj: thanks for catching, it was always putting a break but now I have modified it so that graph will be continuous if the points between end of reference guideset and custom start date are less than 6.

labkey-jeckels (approved):

PR #340, opened 2021-02-25
Item 8387: Reduce memory usage when importing Skyline files

labkey-ankurj:
#### Rationale
During import of large Skyline files into Panorama, much of the document is held in memory which is causing OutOfMemoryErrors for some really large documents. This PR aims to optimize the memory usage during import process.

#### Changes
* Optimize ChromTransition readings
* Reduce in-memory footprint of ChromGroupHeaderInfo

### DIFF
--- a/src/org/labkey/targetedms/parser/SkylineBinaryParser.java
+++ b/src/org/labkey/targetedms/parser/SkylineBinaryParser.java
@@ -17,6 +17,7 @@ package org.labkey.targetedms.parser;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
 import org.labkey.api.exp.api.DataType;
 import org.labkey.api.pipeline.PipelineJobException;
 import org.labkey.targetedms.parser.skyd.CacheFormat;
@@ -210,6 +211,7 @@ public class SkylineBinaryParser
         }
     }
 
+    @NotNull
     private ChromTransition[] parseTransitions(long position, int count) throws IOException
     {
         _channel.position(position);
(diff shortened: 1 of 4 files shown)


src/org/labkey/targetedms/parser/SkylineBinaryParser.java
```diff
@@ -254,26 +253,43 @@ public int matchTransitions(ChromGroupHeaderInfo header, List<? extends GeneralT
                 return match;
         }
 
-        for (GeneralTransition transition : transitions)
+        var start = header.getStartTransitionIndex();
+        var end = start + header.getNumTransitions();
+        var numChromTransitions = end - start + 1;
```
labkey-jeckels:
```suggestion
        var numChromTransitions = header.getNumTransitions();
        var end = start + numChromTransitions;
```

src/org/labkey/targetedms/parser/SkylineBinaryParser.java
```diff
@@ -254,26 +253,43 @@ public int matchTransitions(ChromGroupHeaderInfo header, List<? extends GeneralT
...
+        {
+            _channel.position(_cacheHeaderStruct.getLocationTransitions() + ((long) _cacheHeaderStruct.getChromTransitionSize() * start));
+             chromTransitions = parseTransitions(chromTransitionPosition, numChromTransitions);
+        }
+        catch (IOException e)
+        {
+            throw new RuntimeException(e);
+        }
+
+        if (null != chromTransitions)
```
labkey-jeckels: I don't think chromTransitions can ever be null. If you agree, please annotate parseTransitions() as @NotNull and remove the if()

src/org/labkey/targetedms/parser/skyd/ChromGroupHeaderInfo.java
```diff
@@ -45,52 +45,56 @@
...
     private int uncompressedSize;
-    private Float startTime;
-    private Float endTime;
+    private float startTime;
+    private float endTime;
     private float collisionalCrossSection;
 
     public ChromGroupHeaderInfo(CacheFormatVersion cacheFormatVersion, LittleEndianInput dataInputStream)
     {
+        // For reducing memory usage, following assignments to the above instance variables are commented out because of
```
labkey-jeckels: These variables need to be commented out above to actually reduce the memory footprint.

src/org/labkey/targetedms/parser/skyd/ChromGroupHeaderInfo.java
```diff
@@ -45,52 +45,56 @@
...
+            /*isProcessedScans =*/ dataInputStream.readByte();
+            /*align1 =*/ dataInputStream.readByte();
+            /*statusId =*/ dataInputStream.readShort();
+            /*statusRank =*/ dataInputStream.readShort();
             precursor = dataInputStream.readDouble();
             locationPoints = dataInputStream.readLong();
         }
         if (cacheFormatVersion.compareTo(CacheFormatVersion.Eleven) < 0) {
             uncompressedSize = -1;
+            startTime = Float.MIN_VALUE;
```
labkey-jeckels: Float.MIN_VALUE is a tiny but positive number. I suggest using Float.NEGATIVE_INFINITY instead, and assigning it to a static final variable called NULL_PLACEHOLDER or something like that with a comment that it's used to reduce the memory footprint

src/org/labkey/targetedms/parser/skyd/ChromGroupHeaderInfo.java
```diff
@@ -247,9 +259,6 @@ public boolean isNegativePolarity() {
     }
 
     public boolean excludesTime(double time) {
-        if (null == startTime || null == endTime) {
-            return false;
-        }
         return startTime > time || endTime < time;
```
labkey-jeckels, line 268: This needs to treat the placeholder values as if they were null and return false.

labkey-jeckels (changes requested):

labkey-jeckels (approved):

PR #341, opened 2021-02-26
Progress towards 5x5 reporting in chromatogram libraries

labkey-jeckels:
#### Rationale
We're extending Passport's special reporting to handle 3x3 or 5x5 replicate comparisons

#### Changes
* Overlaid scatter plot, box plot, and CV bar plot for peptides within a protein
* Offer SVG-based plotting for all Panorama JFreeChart rendered plots
* Client side code to offer PDF and PNG downloads of SVG-based plots
* Pull titles and x-axis labels out of plot into surrounding HTML
* Link to replicates list from comparison plots

PR #342, opened 2021-03-01
Switch to use Batik for SVG rendering so we don't need another dependency

labkey-jeckels:
#### Rationale
My recent commit added SVG rendering to JFreeChart plots generated on the server via a new third-party dependency. We can leverage our existing Batik dependency instead, which fixes BasicTest's JAR credits check

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/341

#### Changes
* Switch to Batik SVG rendering
* Adjust XML generation slightly

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -18,6 +18,9 @@ package org.labkey.targetedms;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.keypoint.PngEncoder;
+import org.apache.batik.dom.GenericDOMImplementation;
+import org.apache.batik.svggen.SVGGraphics2D;
+import org.apache.batik.svggen.SVGGraphics2DIOException;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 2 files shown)


labkey-ankurj:
I think the target branch should be release21.3-SS

labkey-jeckels:
I agree but it doesn't exist quite yet

src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -1508,12 +1512,26 @@ else if ("pdf".equalsIgnoreCase(form.getFormat()))
...
+        // Create an instance of the SVG Generator from Batik
+        SVGGraphics2D svgGenerator = new SVGGraphics2D(document);
+        Dimension dim = new Dimension(form.getChartWidth(), form.getChartHeight());
+        svgGenerator.setSVGCanvasSize(dim);
+        Rectangle r = new Rectangle(dim);
+
+        chart.draw(svgGenerator, r);
+
+        // Render SVG to a string
+        StringWriter out = new StringWriter();
```
labkey-ankurj, line 1532: should it be inside try-with-resources? I guess not as it is not actually holding any underlying resources ?

labkey-ankurj (approved):

labkey-danield (approved):

src/org/labkey/targetedms/TargetedMSController.java
labkey-jeckels: Yeah, since it's a StringWriter I didn't bother.

PR #343, opened 2021-03-03
Item 8417: Show reference guideset to inform the selected date range

labkey-ankurj:
#### Rationale
If the date range(s) for the Guide Set(s) are too far apart from the selected dates, data from the dates in between will be truncated so that the Guide Set can be displayed together with the QC data from the custom dates.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2055

#### Changes
* New checkbox option to see reference guide set

### DIFF
--- a/resources/queries/targetedms/GuideSetForOutliers.sql
+++ b/resources/queries/targetedms/GuideSetForOutliers.sql
@@ -2,7 +2,8 @@
 SELECT RowId,
        TrainingStart,
        TrainingEnd,
-       CAST(ReferenceEnd AS TIMESTAMP) AS ReferenceEnd
+       CAST(ReferenceEnd AS TIMESTAMP) AS ReferenceEnd,
+       false as isDefault
 FROM Guideset
 
 UNION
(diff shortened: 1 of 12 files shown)


labkey-jeckels (changes requested):
This change adds 75MB of sample files. I was hoping that we'd be able to edit one of the existing files instead of creating a whole separate copy. And why is one of the .sky.zip files 50MB? The starting file was 25MB.

labkey-sweta:
> This change adds 75MB of sample files. I was hoping that we'd be able to edit one of the existing files instead of creating a whole separate copy. And why is one of the .sky.zip files 50MB? The starting file was 25MB.

Good catch josh, i had accidental zipped extra folder in SProCoPTutorial-QCFolderData.zip. I corrected that.

We created a new copy of new sample file because we were updating the acquired_time in the file and not to mess up other test using this file. In hindsight i should have selected the file which was smaller in size. 

Now only 25 Mb is additional sample file for QC folder.

Let me know if thats needs.

labkey-jeckels (approved):
This is OK for 21.3, but please keep it in mind going forward and maybe we can consolidate files to keep the footprint lean.

You may need to update the tooltip text that contains a filename that's not in use anymore. Good to merge if the test is passing.

PR #353, opened 2021-03-15
Issue 42694: Optimization settings not being pulled from the real source data

labkey-jeckels:
#### Rationale
We want to pull from targetedms.TransitionOptimization, not the CE and DP of the transition itself

#### Changes
* Propagate rows from TransitionOptimization table into the two tables in .clib


labkey-jeckels:
Still need to get good test file(s) for verification

labkey-ankurj (approved):

PR #357, opened 2021-03-17
Item 8607: QC quantitative metric for traces in chromatogram

labkey-ankurj:
#### Rationale
New feature to add a custom metric to set a time point at which to pull the specific trace within the chromatogram.

#### Related Pull Requests
* https://github.com/LabKey/PanoramaPremiumModules/pull/51

#### Changes
* New table - QCTraceMetricValues to store trace values of defined metric
* New columns in QCMetricConfiguration table to support trace metrics


resources/schemas/dbscripts/postgresql/targetedms-21.002-21.003.sql
```diff
@@ -0,0 +1,23 @@
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN TraceValue REAL;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN TimeValue REAL;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN Trace BIGINT;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN YAxisLabel VARCHAR(200);
```
labkey-jeckels: Can we use the existing Series1Label column instead of introducing a separate one just for traces?

resources/schemas/dbscripts/postgresql/targetedms-21.002-21.003.sql
```diff
@@ -0,0 +1,23 @@
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN TraceValue REAL;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN TimeValue REAL;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN Trace BIGINT;
```
labkey-jeckels: As we discussed, this should be the trace's name, not a FK the SampleFileChromInfo.

src/org/labkey/targetedms/TargetedMSManager.java
```diff
@@ -1451,6 +1458,25 @@ public static void purgeDeletedSampleFiles(long sampleFileId)
         // Delete from PeptideChromInfo
         execute("DELETE FROM " + getTableInfoGeneralMoleculeChromInfo() + " WHERE SampleFileId = ?", sampleFileId);
 
+        // Delete from QCTraceMetricValues
+        execute("DELETE FROM " + getTableQCTraceMetricValues() + " WHERE SampleFileId = ?", sampleFileId);
+
+        // Delete from QCEnabledMetrics
+        execute("DELETE FROM " + getTableInfoQCEnabledMetrics() + " WHERE metric IN ( " +
```
labkey-jeckels: Once these are detached from the SampleFileChromInfo, they should go back into TargetedMSListener

src/org/labkey/targetedms/model/QCMetricConfiguration.java
```diff
@@ -31,6 +31,10 @@
     private String _enabledQueryName;
     private String _enabledSchemaName;
     private Boolean _enabled;
+    private Integer _trace;
```
labkey-jeckels: Should be "String _traceName" or something like that

src/org/labkey/targetedms/query/QCMetricConfigurationTable.java
```diff
@@ -75,4 +91,79 @@ public static ContainerFilter getDefaultMetricContainerFilter(Container currentC
...
+        {
+            var insertedRow = super.insertRow(user, container, row);
+            calculateAndInsertTraceValuesForMetric((int) insertedRow.get("Id"), container, user);
+            return insertedRow;
+        }
+
+        @Override
+        public List<Map<String, Object>> insertRows(User user, Container container, List<Map<String, Object>> rows, BatchValidationException errors, @Nullable Map<Enum, Object> configParameters, Map<String, Object> extraScriptContext) throws DuplicateKeyException, QueryUpdateServiceException, SQLException
+        {
+            return super.insertRows(user, container, rows, errors, configParameters, extraScriptContext);
```
labkey-jeckels: Do we need this override?

src/org/labkey/targetedms/query/QCMetricConfigurationTable.java
```diff
@@ -75,4 +91,79 @@ public static ContainerFilter getDefaultMetricContainerFilter(Container currentC
...
+    {
+        protected QCMetricConfigurationTableUpdateService(TableInfo queryTable, TableInfo realTable)
+        {
+            super(queryTable, realTable);
+        }
+
+        @Override
+        protected Map<String, Object> getRow(User user, Container container, Map<String, Object> keys) throws InvalidKeyException, QueryUpdateServiceException, SQLException
+        {
+            return super.getRow(user, container, keys);
```
labkey-jeckels: Do we need this override?

src/org/labkey/targetedms/query/QCMetricConfigurationTable.java
```diff
@@ -75,4 +91,79 @@ public static ContainerFilter getDefaultMetricContainerFilter(Container currentC
...
+        protected Map<String, Object> deleteRow(User user, Container container, Map<String, Object> oldRow) throws InvalidKeyException, QueryUpdateServiceException, SQLException
+        {
+            deleteTraceValueForMetric((Integer) oldRow.get("id"));
+            return super.deleteRow(user, container, oldRow);
+        }
+
+        @Override
+        public List<Map<String, Object>> deleteRows(User user, Container container, List<Map<String, Object>> keys, @Nullable Map<Enum, Object> configParameters, @Nullable Map<String, Object> extraScriptContext) throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
+        {
+            return super.deleteRows(user, container, keys, configParameters, extraScriptContext);
```
labkey-jeckels: Same here - can this be deleted?

src/org/labkey/targetedms/query/QCMetricConfigurationTable.java
```diff
@@ -75,4 +91,79 @@ public static ContainerFilter getDefaultMetricContainerFilter(Container currentC
...
+        }
+
+        private void deleteTraceValueForMetric(int metricId)
+        {
+            var sql = new SQLFragment("DELETE FROM " + TargetedMSManager.getTableQCTraceMetricValues() +
```
labkey-jeckels: Add a container argument and make this join through the SampleFile table to prevent an attacker that has delete access in one container being able to delete the metric values from a container where they don't have permission

labkey-jeckels (changes requested):

resources/schemas/dbscripts/postgresql/targetedms-21.002-21.003.sql
labkey-ankurj: Yes

resources/schemas/dbscripts/postgresql/targetedms-21.002-21.003.sql
labkey-ankurj: OK

src/org/labkey/targetedms/model/QCMetricConfiguration.java
labkey-ankurj: OK

src/org/labkey/targetedms/query/QCMetricConfigurationTable.java
labkey-ankurj: nope

src/org/labkey/targetedms/query/QCMetricConfigurationTable.java
labkey-ankurj: nope

src/org/labkey/targetedms/query/QCMetricConfigurationTable.java
labkey-ankurj: sure

src/org/labkey/targetedms/TargetedMSManager.java
labkey-ankurj: done

resources/schemas/dbscripts/sqlserver/targetedms-21.002-21.003.sql
```diff
@@ -1,11 +1,6 @@
 ALTER TABLE targetedms.QCMetricConfiguration ADD TraceValue REAL;
 ALTER TABLE targetedms.QCMetricConfiguration ADD TimeValue REAL;
-ALTER TABLE targetedms.QCMetricConfiguration ADD Trace BIGINT;
-ALTER TABLE targetedms.QCMetricConfiguration ADD YAxisLabel VARCHAR(200);
-
-ALTER TABLE targetedms.QCMetricConfiguration
-    ADD CONSTRAINT FK_QCMetricConfiguration_Trace
-        FOREIGN KEY (Trace) REFERENCES targetedms.SampleFileChromInfo(Id)
+ALTER TABLE targetedms.QCMetricConfiguration ADD TraceName VARCHAR(200);
```
labkey-jeckels: Should be NVARCHAR for SQLServer

labkey-jeckels (approved):
One small followup tweak, but otherwise all set to merge.

resources/queries/targetedms/QCTraceMetric.sql
```diff
@@ -0,0 +1,2 @@
+SELECT 0 AS PrecursorChromInfoId, SampleFileId, metric.Name AS SeriesLabel, value as MetricValue, metric
```
labkey-jeckels: Please add a comment explaining how this query fits in. It's special and can't be used as-is. It needs the help of the Java code

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -247,19 +257,20 @@ private String queryContainerSampleFileRawData(List<QCMetricConfiguration> confi
     public String getMetricLabel(Map<Integer, QCMetricConfiguration> metrics, RawMetricDataSet dataRow)
     {
         QCMetricConfiguration metric = metrics.get(dataRow.getMetricId());
-        String metricLabel;
-        switch (dataRow.getMetricSeriesIndex())
+
+        if (metric.getTraceName() != null)
+        {
+            return metric.getName();
```
labkey-jeckels: Can we store the value we want in the series1Label column so we don't need to special case the code here?

resources/queries/targetedms/QCTraceMetric.sql
labkey-jeckels: though I'm still not clear why we need this query file at all. Could we in-line this in the Java code that's building the query, so that instead of tacking on the "WHERE metric=x" it appends the whole thing? This approach seems like we have more pieces in different places, some of which aren't useful on their own

labkey-jeckels (approved):
A few ideas to consider though.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-ankurj: I am storing YAxisLabel in series1Label column.

resources/queries/targetedms/QCTraceMetric.sql
labkey-ankurj: I can definitely add a comment in this query. It is similar to other metric queries like TicArea  and we have to store the queryName in QCMetricConfiguration for the trace metrics and if we do in-line java approach, it would require a similar check to write a custom select (Select 0 as PrecursorChromInfoId) for trace metrics.

labkey-ankurj: Also, I am not clear on how other metric queries are different than this one in terms of using them.

webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -779,7 +750,7 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
                     color: '#555555'
                 },
                 yLeft: {
-                    value: this.getYScaleLabel(plotType, trendLineProps.valueConversion, metricProps.series1Label),
+                    value: this.getYScaleLabel(plotType, trendLineProps.valueConversion, metricProps.yAxisLabel1),
```
labkey-jeckels, line 753: Wouldn't we need to reference yAxisLabel2 somewhere in the JS?

labkey-jeckels (changes requested):

webapp/TargetedMS/js/QCPlotHelperBase.js
labkey-ankurj: yes, added that in yRight

labkey-jeckels (approved):

PR #370, opened 2021-04-29
Issue 42839: Link to QC folder from experimental folder doesn't zoom date range

labkey-ankurj:
#### Rationale
* When coming from experiment folder to QC folder - 
   If reference guideset is present - display 5 points before the experimental folder start date and reference guideset . Date range filter will have the start date as the date of acquired time of 1st point out of five points. We will auto-check the box to include the reference guide set.
  if reference guideset is not present - still display the 5 points before the experimental folder start date . Date range filter will have the start date as the date of acquired time of 1st point out of five points that will be shown.
 End date in both cases will be the end date of experimental folder.
 Experiment range toolbar will show the experimental folder start and end date without the timestamp.

* When filtering the dates in QC folder - do not display the 5 points before the selected start date but will show the reference guideset if present.

### DIFF
--- a/webapp/TargetedMS/js/BaseQCPlotPanel.js
+++ b/webapp/TargetedMS/js/BaseQCPlotPanel.js
@@ -203,14 +203,14 @@ Ext4.define('LABKEY.targetedms.BaseQCPlotPanel', {
     queryQCInstruments: function(successCallback, callbackScope) {
         LABKEY.Query.executeSql({
             schemaName: 'targetedms',
-            sql: 'SELECT DISTINCT instrumentSerialNumber FROM samplefile',
+            sql: 'SELECT DISTINCT instrumentSerialNumber, instrumentId.model FROM samplefile',
             containerFilter: LABKEY.Query.containerFilter.current,
             scope: this,
             success: function (response) {
 
                 if (response.rows && response.rows.length > 0) {
                     this.qcIntrumentsArr = response.rows.map(function (row) {
-                        return row.instrumentSerialNumber;
+                        return row.instrumentSerialNumber + (row.model ? ' - ' + row.model : '');
                     })
                 }
 
--- a/webapp/TargetedMS/js/BaseQCPlotPanel.js
+++ b/webapp/TargetedMS/js/BaseQCPlotPanel.js
@@ -210,7 +210,15 @@ Ext4.define('LABKEY.targetedms.BaseQCPlotPanel', {
 
                 if (response.rows && response.rows.length > 0) {
                     this.qcIntrumentsArr = response.rows.map(function (row) {
-                        return row.instrumentSerialNumber + (row.model ? ' - ' + row.model : '');
+                        if (row.instrumentSerialNumber && row.model) {
+                            return row.instrumentSerialNumber + ' - ' + row.model;
+                        }
+                        else if (row.instrumentSerialNumber && !row.model) {
+                            return row.instrumentSerialNumber;
+                        }
+                        else if (!row.instrumentSerialNumber && row.model) {
+                            return row.model;
+                        }
                     })
                 }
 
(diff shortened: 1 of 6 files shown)


webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -196,6 +185,47 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
+        var sql = "Select MIN(sf.AcquiredTime) AS StartDate,\n" +
+                "       MAX(sf.AcquiredTime) AS EndDate,\n" +
+                "       r.FileName,\n" +
+                "       sf.InstrumentSerialNumber,\n" +
+                "       sf.InstrumentId.model\n" +
+                "FROM targetedms.Replicate rep\n" +
+                "INNER JOIN targetedms.SampleFile sf ON rep.Id = sf.ReplicateId\n" +
+                "INNER JOIN targetedms.Runs r on r.Id = rep.RunId\n" +
+                "where rep.RunId ='" + runId + "'\n" +
+                "GROUP BY r.FileName, sf.InstrumentSerialNumber, sf.InstrumentId.model";
```
labkey-jeckels: This grouping approach won't work if the file in question has data from multiple instruments. We'll be grouping to separate them into different rows, but then we only use the first one (which may have nothing to do with this QC folder) in the success handler.

What are we doing with the serial number and model in the client side code right now?

Also, given the way that we generate the SQL behind the scenes for allFolders queries, we'd get better perf if you just do the SampleFile table in your FROM clause and use the lookup syntax to get the replicate and run columns.

labkey-jeckels (changes requested):

webapp/TargetedMS/js/QCTrendPlotPanel.js
labkey-ankurj: I will update the sql. 

Serial Number and model (instrument name) are used in the hover tooltip of experiment range highlighted region in QC Plots.

labkey-jeckels: Hmm... let's remove it from the tooltip and simplify the query so that it's no longer grouping or returning those columns. That will eliminate the problem of choosing the wrong instrument's data if there are multiple in the referenced Skyline document.

This will be better, though we should someday filter it for just the data that matches instrument in the QC folder.

webapp/TargetedMS/js/QCSummaryPanel.js
```diff
@@ -48,13 +50,19 @@ Ext4.define('LABKEY.targetedms.QCSummary', {
                 container.showName = hasChildren;
                 container.isParent = true;
                 container.parentOnly = containers.length == 1;
+                if (this.qcPlotPanel.qcIntrumentsArr) {
+                    if (this.qcPlotPanel.qcIntrumentsArr.length > 1) {
+                        container.instrument = 'multiple instruments. QC folders should have data from only one instrument.'
```
labkey-jeckels:
```suggestion
                        container.instrument = 'multiple instruments. We recommend that each instrument use its own QC folder.'
```

labkey-jeckels: Can we show a comma-separated list of instruments here?

webapp/TargetedMS/js/BaseQCPlotPanel.js
```diff
@@ -200,6 +200,26 @@ Ext4.define('LABKEY.targetedms.BaseQCPlotPanel', {
         });
     },
 
+    queryQCInstruments: function(successCallback, callbackScope) {
+        LABKEY.Query.executeSql({
+            schemaName: 'targetedms',
+            sql: 'SELECT DISTINCT instrumentSerialNumber FROM samplefile',
```
labkey-jeckels: Let's include the model too, since people are more likely to recognize 'Q Exactive' than a potentially obscure serial number.

labkey-jeckels (approved):
A couple more small tweaks but looks good. No further review needed unless you have questions

PR #373, opened 2021-05-02
Issue 43061: Ignore zero byte spectrum library files

vagisha:
#### Rationale
Some folders on PanoramaWeb have zero byte spectrum library (.blib) files.  It is not yet clear how they got created.  They are not included in the .sky.zip files even though the spectrum library was listed in the Skyline XML settings. Attempting to read these files throws an exception. We will ignore zero byte files. The code that is creating them will be fixed once we know what it is.


labkey-jeckels (approved):

vagisha:
This PR also fixed the problem of 0-byte .blib files. More details are in the issue tracker: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43061

PR #377, opened 2021-05-06
Issue 42754: Narrowing down date range in QC plots changes the y axis scales due to which mean and SD lines in QC plots are removed

labkey-ankurj:
#### Rationale
Narrowing down date range in QC plots changes the y axis scales due to which mean and SD lines in QC plots are removed. This PR fixes the issue.

#### Changes
* if the date range is zoomed, send correct stats


labkey-ankurj:
@labkey-jeckels I tested with Log statements for checking the performance, it was within milli seconds so I didn't take the approach of filtering the raw qc plots data in java.

labkey-jeckels:
@labkey-ankurj How long was it taking? I'm worried that it's fast enough when there are 20 samples that the difference seems negligible, but in a real-world scenario of having 1,000 samples it'll be more painful. For example, see this folder:

https://panoramaweb.org/project/home/Example%20Data/Quality%20Control%202/begin.view?

src/org/labkey/targetedms/TargetedMSManager.java
```diff
@@ -2571,4 +2571,15 @@ else if (traceValue != null && values[i] >= traceValue)
                 .filter(qcMetricConfiguration -> qcMetricConfiguration.getTraceName() != null)
                 .collect(Collectors.toList());
     }
+
+    public static Map<String,Object> getQCFolderDateRange(Container container)
```
labkey-jeckels, line 2594: What about having this return a Pair<Date, Date> instead? That would eliminate the need to cast and know the keys to the map.

src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -1293,14 +1293,37 @@ public Object execute(QCPlotsDataForm form, BindException errors) throws Excepti
...
+
             List<RawMetricDataSet> rawMetricDataSets = generator.getRawMetricDataSets(getContainer(), getUser(), qcMetricConfigurations, rangeStartDate, form.getEndDate(), form.getSelectedAnnotations(), form.isShowExcluded());
             Map<GuideSetKey, GuideSetStats> stats = generator.getAllProcessedMetricGuideSets(rawMetricDataSets, guideSets.stream().collect(Collectors.toMap(GuideSet::getRowId, Function.identity())));
+
+            // get start date and end date for this qc folder
+            Map<String,Object> qcFolderDateRange = TargetedMSManager.getQCFolderDateRange(getContainer());
+            Date qcStartDate = (Date) qcFolderDateRange.get("startDate");
+            Date qcEndDate = (Date) qcFolderDateRange.get("endDate");
+
+            boolean zoomedRange = qcStartDate.compareTo(rangeStartDate) != 0 || qcEndDate.compareTo(form.getEndDate()) != 0;
```
labkey-jeckels: Will we never get to this code if the folder is empty? I'm wondering if the start and end dates could ever be null.

labkey-jeckels (approved):
Suggestions but no need for additional review unless you have questions or concerns

labkey-ankurj:
Screen Shot 2021-05-06 at 2 06 05 PM
1 - start querying the targeted qc plots data (before L1302)
2 - got targeted qc plots data (after L1303)
3- start querying the full qc plots data (before L1315)
4 - got full qc plots data (after L1316)

labkey-jeckels:
@labkey-ankurj I tested with the 1000+ sample folder and it's taking around 1.1 seconds to run that query. I think it's worth spending a little more time to try avoiding double-querying. The page is already taking longer than we'd like to render on those big datasets

src/org/labkey/targetedms/TargetedMSController.java
labkey-jeckels: @labkey-ankurj did you see this comment about a possible NPE?

PR #378, opened 2021-05-11
Issue 43099: Add new usage metric for guidesets in Panorama QC plots

labkey-jeckels:
#### Rationale
We're curious about real-world guide set usage in Panorama

#### Changes
* Capture the total number of guide sets defined on a server and the number of containers that have them defined
* Also capture folder type breakdown


labkey-ankurj (approved):

labkey-jeckels:
Added another set of metrics, @labkey-ankurj can you review again?

labkey-ankurj (approved):
looks good.

PR #383, opened 2021-05-14
Import all TransitionChromInfos when there are < 1000 precursors

labkey-jeckels:
#### Rationale
We recently instituted a cutoff so that we don't import TransitionChromInfos for very large documents. We want to be more selective about applying the cutoff

#### Changes
* Only skip importing TransitionChromInfos if there are more than 100,000 AND there are at least 1,000 precursors

### DIFF
--- a/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
+++ b/src/org/labkey/targetedms/parser/SkylineDocumentParser.java
@@ -183,7 +183,8 @@ public class SkylineDocumentParser implements AutoCloseable
 
     /**
      * To prevent giant DIA documents from overwhelming the DB, we skip importing TransitionChromInfos if the document
-     * has more than 100,000 AND has more than 1,000 precursors
+     * has more than 100,000 AND has more than 1,000 precursors. We use both because a document may have a lot of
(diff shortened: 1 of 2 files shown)


src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -183,9 +183,10 @@
 
     /**
      * To prevent giant DIA documents from overwhelming the DB, we skip importing TransitionChromInfos if the document
-     * has more than 100,000
+     * has more than 100,000 AND has more than 1,000 precursors
```
labkey-ankurj:
```suggestion
     * has more than 100,000 OR has more than 1,000 precursors
```

labkey-ankurj (approved):

vagisha (approved):

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
labkey-jeckels: The text is accurate. It's describing the opposite behavior of the method below, hence the < comparison and || in the code.

PR #391, opened 2021-06-09
Assorted small exception report fixes

labkey-jeckels:
#### Rationale
Issue 43328: NullPointerException in org.labkey.targetedms.chart.ComparisonChartMaker.makeRetentionTimesChart()

Issue 43326: Error handling improvements when SKYD files aren't available on disk

Issue 43325: IllegalStateException from org.labkey.targetedms.outliers.OutlierGenerator.getSampleFiles()

#### Changes
* Demote file system problem to logging or ConfigurationException to avoid reporting
* Fix NPEs
* Collapse duplicate SampleFile entries from query

### DIFF
--- a/src/org/labkey/targetedms/chart/ChromatogramDataset.java
+++ b/src/org/labkey/targetedms/chart/ChromatogramDataset.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.targetedms.chart;
 
+import org.apache.log4j.Logger;
 import org.jfree.chart.ChartColor;
 import org.jfree.data.xy.XYSeries;
 import org.jfree.data.xy.XYSeriesCollection;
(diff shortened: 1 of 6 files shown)


src/org/labkey/targetedms/chart/ChromatogramDataset.java
```diff
@@ -65,6 +66,8 @@
  */
 public abstract class ChromatogramDataset
 {
+    private static final Logger LOG = Logger.getLogger(ChromatogramDataset.class);
```
labkey-ankurj:
```suggestion
    private static final Logger LOG = LogManager.getLogger(ChromatogramDataset.class);
```

labkey-ankurj: this will require changing imports.

labkey-ankurj (changes requested):

labkey-ankurj (approved):

PR #396, opened 2021-06-24
Fix calibration curves for large documents (when TransitionChromInfos do not get saved in database)

nickshulman:
#### Rationale
Calibration curves and fold changes are currently unavailable in Panorama for Skyline documents which have more than 1000 precursors and more than 100,000 transition results (see "SkylineDocumentParser.shouldSaveTransitionChromInfos")
This causes a problem for calibration curves and fold changes, since those are currently calculated using the TransitionChromInfo table.
This change adds some extra information to PrecursorChromInfo table which can be used to calculate these things if there is nothing in the TransitionChromInfo table.

#### Changes
Added new columns "TotalAreaMs1" and "TotalAreaFragment" to PrecursorChromInfo. These get calculated from TransitionChromInfo objects in "SkylineDocumentParser.computePrecursorChromInfoValues".
Added "TrnasitionManager.runHasTransitionChromInfos" which checks whether there are any TransitionChromInfo rows associated with the run.
If the run does not have any TransitionChromInfos, then GeneralMoleculeResultDataSet uses the precursor-level numbers instead of the transition level numbers.


src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -1761,56 +1763,99 @@ else if (XmlUtil.isStartElement(reader, evtType, NOTE))
...
+     * Also: populate TotalAreaMs1 and TotalAreaFragment
+     *
+     * @param precursor
+     */
+    private void computePrecursorChromInfoValues(GeneralPrecursor<?> precursor)
     {
         TransitionSettings.FullScanSettings fullScanSettings = _transitionSettings == null ? null : _transitionSettings.getFullScanSettings();
 
         // Key is the replicate name
         // Value is a List where each member is a Pair of TransitionChromInfo and a Boolean value indicating if the transition is quantitative
```
labkey-jeckels, line 1817:
```suggestion
        // Value is a List where each member is a Tuple of TransitionChromInfo and a Boolean value indicating if the transition is quantitative, and the msLevel
```

labkey-jeckels (approved):

labkey-jeckels:
I've been thinking about testing for these scenarios as we continue to add handling for these large documents. Two ideas:

1. Have a secret action that deletes all of the TransitionChromInfos for a given Skyline document after it's been imported. This is easy, though the import code wouldn't go through the large document handling.
2. Make the thresholds configurable (perhaps via a module property) so that the tests (or a real-world admin) could adjust it from its default value.

Any preferences?

labkey-jeckels:
Also, it looks like your test run got unlucky and happened to fire when buttons were briefly broken. I triggered new runs on both PG and SQLServer

nickshulman:
Making those cutoff numbers configurable by an admin sounds like good feature. Let's do that.

The calibration curves get computed as part of importing the document. If you were to delete the TransitionChromInfo's after the fact, that would be too late to exercise the calibration curve code.


> I've been thinking about testing for these scenarios as we continue to add handling for these large documents. Two ideas:
> 
> 1. Have a secret action that deletes all of the TransitionChromInfos for a given Skyline document after it's been imported. This is easy, though the import code wouldn't go through the large document handling.
> 2. Make the thresholds configurable (perhaps via a module property) so that the tests (or a real-world admin) could adjust it from its default value.
> 
> Any preferences?

labkey-jeckels:
> Making those cutoff numbers configurable by an admin sounds like good feature. Let's do that.

OK, I'm working on this change. I'm planning to push the changes into this FB.

labkey-jeckels:
I pushed the change to make the limits configurable. You can manually adjust them by going to Folder->Management->Module Properties, or in test code. Here's an example:

https://github.com/LabKey/targetedms/blob/33b3684477eb225b16e680401cbce56694763fb8/test/src/org/labkey/test/tests/targetedms/TargetedMSQCSummaryTest.java#L109

PR #398, opened 2021-07-04
Issue 43512: Expand FileNameHint and AminoAcid columns

vagisha:
#### Rationale
Users reported failed Skyline document imports due to insufficient limits on the following columns:
SpectrumLibrary.FileNameHint
and
StructuralModification.AminoAcid
IsotopeModificaiton.AminoAcid


labkey-jeckels (approved):
This looks fine. We've already branched for 21.7 so this will end up in 21.11 as currently set up. Of course, you can run the script manually against PanoramaWeb whenever you like, so it's up to you on where to merge it. I doubt other installs will hit the current limit.

PR #399, opened 2021-07-11
Support chromatogram libraries in Panorama Public

vagisha:
#### Rationale
We want to allow submission of chromatogram library folders to Panorama Public. Here we are extending the targetedms api to support the code that will be added to the panoramapublic module.


api-src/org/labkey/api/targetedms/ITargetedMSRun.java
```diff
@@ -26,13 +26,34 @@
...
-    public String getBaseName();
-    public String getFileName();
-    public String getDescription();
-    public Date getCreated();
-    public long getId();
-    public Integer getDataId();
-    public Integer getSkydDataId();
-    public String getSoftwareVersion();
+    /** Don't change the ordering of these enum values without updating the values in targetedms.runs.representativedatastate */
+    enum RepresentativeDataState
```
labkey-jeckels: I know this predates the move, but I think we should rename one of the two enums named RepresentativeDataState. Maybe this one should be RunRepresentativeDataState?

labkey-jeckels (approved):

labkey-jeckels (approved):

PR #400, opened 2021-07-12
Automation Test

labkey-sweta:
#### Rationale
 Reproducibility Report Enhancements automation test
1. Reproducibility Report test case
2. Calibration curve test case
3. Clib File test case.

#### Related Pull Requests
* 

#### Changes
additional test cases in test/src/org/labkey/test/tests/targetedms/TargetedMSMxNReproducibilityReportTest.java

### DIFF
--- a/test/src/org/labkey/test/tests/targetedms/TargetedMSMxNReproducibilityReportTest.java
+++ b/test/src/org/labkey/test/tests/targetedms/TargetedMSMxNReproducibilityReportTest.java
@@ -132,6 +132,8 @@ public class TargetedMSMxNReproducibilityReportTest extends TargetedMSTest
         clickAndWait(calibrationCurvesLink().index(2));
 
         log("Verifying the FOM values");
+        waitForElement(Locator.css("span.labkey-wp-title-text").withText("Figures of Merit"));
+        checker().verifyTrue("Incorrect peptide", isElementPresent(Locator.tagWithText("h3", "Calibration Curve: VIFDANAPVAVR")));
         checker().verifyEquals("Incorrect Lower limit of quantitation", "1.0",
                 Locator.tagWithId("td", "lloq-stat").findElement(getDriver()).getText());
         checker().verifyEquals("Incorrect Upper limit of quantitation", "10.0",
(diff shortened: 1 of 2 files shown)


test/src/org/labkey/test/tests/targetedms/TargetedMSMxNReproducibilityReportTest.java
```diff
@@ -93,6 +114,75 @@ public void testSingleReplicateFile()
...
+        importData(SKY_FILE_FOR_CALIBRATION_CURVE);
+
+        clickTab("Proteins");
+
+        checker().verifyFalse("Reproducibility report link should not be present", isElementPresent(reproducibilityReportLink()));
+        checker().verifyTrue("Calibration curve icon should be present", isElementPresent(calibrationCurvesLink()));
+
+        clickAndWait(calibrationCurvesLink().index(2));
+
+        log("Verifying the FOM values");
```
labkey-jeckels, line 134: If we're just clicking on the third link (and not identifying it based on the protein name) let's also verify that we're getting to the peptide we're expecting, along with the upper and lower limits

test/src/org/labkey/test/tests/targetedms/TargetedMSMxNReproducibilityReportTest.java
```diff
@@ -93,6 +114,75 @@ public void testSingleReplicateFile()
...
+        checker().verifyEquals("Incorrect Lower limit of quantitation", "1.0",
```
labkey-jeckels, line 137: Might need to wait for these values - they're loaded via AJAX so they may lag the initial page load

labkey-jeckels (approved):
A couple of suggestions but no need for an additional review unless you'd like one.

test/src/org/labkey/test/tests/targetedms/TargetedMSMxNReproducibilityReportTest.java
labkey-sweta: In the Protein tab, Protein/Label column  as it listed as TG|1578-1589|z2 upon clicking reproducibility report link it navigates to page with title "Calibration Curve: VIFDANAPVAVR"

Verify for sequence "VIFDANAPVAVR". ?

labkey-jeckels: Yes, sounds good.

PR #408, opened 2021-07-26
Do not allow Skyline zip files containing more than one .sky file

vagisha:
#### Rationale
One PanoramaWeb user recently uploaded and imported a manually created .zip file containing all their Skyline documents.  Only the first .sky in the zip archive was imported into the database.  Zip files containing multiple .sky files should cause an error.

### DIFF
--- a/src/org/labkey/targetedms/SkylineDocImporter.java
+++ b/src/org/labkey/targetedms/SkylineDocImporter.java
@@ -950,8 +950,14 @@ public class SkylineDocImporter
             for(File file: files)
             {
                 ext = FileUtil.getExtension(file.getName());
-                if(SkylineFileUtils.EXT_SKY.equalsIgnoreCase(ext) && null == skyFile)
+                if(SkylineFileUtils.EXT_SKY.equalsIgnoreCase(ext))
                 {
+                    if (null != skyFile)
+                    {
+                        throw new IOException("zip file " + f + " contains more than one .sky file and is not a valid Skyline document archive. " +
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

labkey-jeckels:
Looks like one of our test files, test/sampledata/TargetedMS/SProCoPTutorial.zip, was manually manipulated in OSX which helpfully injected some hidden files. I don't expect this will be a real-world problem, but we'll need to clean up that .zip before merging

vagisha:
Yes, I just saw that.  There is one more that was also created on a Mac and has the extra __MACOSX folder with a .sky file - SProCoPTutorial-QCFolderData.zip.  I can clean it up.  But I was wondering if we need three copies of the SProCoP Skyline document?  Each one is ~25MB.   The third one is SProCoPTutorial_withAnnotations.zip.
The documents in SProCoPTutorial.zip and SProCoPTutorial-QCFolderData.zip are different only in the acquired and modification times of some of the sample files.  Perhaps one of the files was hand edited for a test?  But I don't see a note about that in TargetedMSExperimentalQCLinkTest.  SProCoPTutorial_withAnnotations.zip has added replicate annotations and several other differences throughout the XML.  
For this PR I will just clean up the two zip files.

labkey-jeckels:
I've tried to keep the duplicates of those big files to a minimum. I think at least one is used to test that we update annotations upon reimport in a QC folder. But we could probably get by with fewer replicates in some of them. Regardless, the minimal fix for this PR seems best.

PR #422, opened 2021-08-26
Look up from Panorama to sample type if we find matching sample IDs

labkey-jeckels:
#### Rationale
Users want to link data from their Skyline documents to sample metadata, especially in LKSM

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2569

#### Changes
* Make sample name a lookup to the sample type (when matches are present) from the listing on the replicates page
* On sample file details page give a link as well

### DIFF
--- a/src/org/labkey/targetedms/query/SampleFileTable.java
+++ b/src/org/labkey/targetedms/query/SampleFileTable.java
@@ -22,18 +22,28 @@ import org.labkey.api.data.Aggregate;
 import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
+import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DataColumn;
 import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.LazyForeignKey;
 import org.labkey.api.data.RenderContext;
 import org.labkey.api.data.SQLFragment;
+import org.labkey.api.data.SqlSelector;
+import org.labkey.api.data.Table;
 import org.labkey.api.data.TableSelector;
 import org.labkey.api.data.dialect.SqlDialect;
 import org.labkey.api.exp.api.ExpData;
+import org.labkey.api.exp.api.ExpSampleType;
+import org.labkey.api.exp.api.ExperimentService;
+import org.labkey.api.exp.api.SampleTypeService;
+import org.labkey.api.exp.query.SamplesSchema;
 import org.labkey.api.query.DefaultQueryUpdateService;
 import org.labkey.api.query.DetailsURL;
 import org.labkey.api.query.ExprColumn;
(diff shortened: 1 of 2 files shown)


src/org/labkey/targetedms/query/SampleFileTable.java
```diff
@@ -97,6 +109,47 @@ public SampleFileTable(TargetedMSSchema schema, ContainerFilter cf, @Nullable Ta
...
+            sql.append(ExperimentService.get().getTinfoMaterial(), "m");
+            sql.append(" WHERE CpasType IS NOT NULL AND Name IN (SELECT SampleName FROM (");
+            Set<ColumnInfo> selectCols = Set.of(getColumn("SampleName"));
+            sql.append(QueryService.get().getSelectSQL(SampleFileTable.this, selectCols, null, null, Table.ALL_ROWS, 0, false));
+            sql.append(") X)");
+
+            Set<Container> matchingContainers = new HashSet<>();
+            Set<String> matchingSampleTypeLSIDs = new HashSet<>();
+
+            // Iterate and make that the user has permission to the target container
```
labkey-susanh:
```suggestion
            // Iterate and make sure the user has permission to the target container
```

src/org/labkey/targetedms/query/SampleFileTable.java
```diff
@@ -97,6 +109,47 @@ public SampleFileTable(TargetedMSSchema schema, ContainerFilter cf, @Nullable Ta
...
+                Container c = ContainerManager.getForId((String)match.get("Container"));
+                if (c != null && c.hasPermission(getUserSchema().getUser(), ReadPermission.class))
+                {
+                    matchingContainers.add(c);
+                    String sampleTypeLSID = (String)match.get("CpasType");
+                    matchingSampleTypeLSIDs.add(sampleTypeLSID);
+                }
+            }
+
+            if (matchingContainers.size() == 1 && matchingSampleTypeLSIDs.size() == 1)
```
labkey-susanh, line 139: How often will it happen that there will be more than one match here?  Is that an error condition that should be logged or just a known instance where you won't get a lookup?

labkey-susanh (approved):

src/org/labkey/targetedms/query/SampleFileTable.java
labkey-jeckels: In most cases I expect it will find 0 or 1 matches. However, on a big server (like PanoramaWeb, used by many labs) there's always a chance that two labs will use overlapping sample IDs in different folders. I wouldn't consider this to be an error condition, but it is unclear which is the "right" one to use, especially for site admins and others with wide-ranging access.

In the future, we may want to prefer matches in the same project, or make it configurable so that you can explicitly identify the container(s) with samples that you want to be in scope. But this seemed like a reasonable starting point that requires zero setup by admins.

PR #429, opened 2021-09-13
Issue 43787: Add use-case guidance to Loggers page of admin console

labkey-jeckels:
#### Rationale
Users like to know what loggers may have debug logging of interest. They can currently find some guidance at https://www.labkey.org/Documentation/wiki-page.view?name=loggers

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2618

#### Changes
* Add some notes about loggers

### DIFF
--- a/src/org/labkey/targetedms/parser/PrecursorChromInfo.java
+++ b/src/org/labkey/targetedms/parser/PrecursorChromInfo.java
@@ -16,12 +16,12 @@
 package org.labkey.targetedms.parser;
 
 import org.apache.commons.io.output.ByteArrayOutputStream;
-import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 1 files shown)


src/org/labkey/targetedms/parser/PrecursorChromInfo.java
```diff
@@ -39,7 +39,7 @@
  */
 public class PrecursorChromInfo extends AbstractChromInfo implements Comparable<PrecursorChromInfo>
 {
-    private static final Logger LOG = LogManager.getLogger(PrecursorChromInfo.class);
+    private static final Logger LOG = LogHelper.getLogger(PrecursorChromInfo.class, "Messages about fetching chromatograms, either from the cache-in memory, or from S3/local files.");
```
labkey-susanh: cache-in memory? What is that?  Is that "cache (in memory)" or "in-memory cache"?

labkey-susanh (approved):

PR #431, opened 2021-09-23
Smarter handling for transition area QC metrics

labkey-jeckels:
#### Rationale
Transition-level data isn't always quantitative. When it's not, it doesn't make sense to use it as a QC metric. Also, users want to be able to visualize precursor and transition peak areas independently, or together

#### Changes
* Enable transition-level area metrics only when we have quantitative data
* Expose transition and precursor areas as their own top-level metrics so they can get the full Levey-Jennings treatment
* Rename the "Peak Area" metric to make it clear that it's the combination of precursor and transition areas


resources/queries/targetedms/QCMetricEnabled_transitionArea.sql
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+-- Approximate the checks in GeneralTransition.isQuantitative()
+SELECT Id, FragmentType, Quantitative FROM Transition t
+    WHERE
+        (Quantitative = TRUE) OR (Quantitative IS NULL AND
+                                  (FragmentType != 'precursor' AND
+                                  t.GeneralPrecursorId.GeneralMoleculeId.PeptideGroupId.RunId IN (
```
ankurjuneja, line 21: is this indentation intentional?

ankurjuneja (approved):

resources/queries/targetedms/QCMetricEnabled_transitionArea.sql
labkey-jeckels: Kind of. IntelliJ has strange ideas for how to indent this SQL and I got tired of fighting it.

PR #432, opened 2021-09-24
Automation test for Add new iRT QC Metric

labkey-sweta:
#### Rationale
Automation test for Add new iRT QC Metric

#### Related Pull Requests
https://github.com/LabKey/PanoramaPremiumModules/pull/58

#### Changes
Adding additional metric values.


labkey-jeckels (changes requested):
Adding these to the enum is fine, but there's now a failure in TargetedMSQCTest.testQCPlotInputs because it's expecting the iRT metrics to show up, but presumably the files don't have iRT data

test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
```diff
@@ -748,6 +747,9 @@ public String getLabel()
         TPAREARATIO("Transition/Precursor Area Ratio", true),
         TPAREAS("Transition & Precursor Areas", true),
         MASSACCURACY("Mass Accuracy", true),
+        IRTINTERCEPT("iRT Intercept", false),
+        IRTSLOPE("iRT Slope", false),
+        IRTCORRELATION("iRT Correlation", false),
```
labkey-jeckels: The "hasData" flag seems like a bad thing to put into what's supposed to be a reusable component. The iRT metrics have data in some (but certainly not all) sample files. It seems like the test, which controls which file(s) have been loaded should be responsible for knowing what metrics should have data.

labkey-jeckels (changes requested):

test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
labkey-sweta: how about i add one more parameter isOptionalMetric and use that in test? or u think tests should handle that?

labkey-jeckels: It looks like the only place that uses hasData() right now is TargetedMSQCTest.testQCPlotInputs(). It calls it four times but it'll never enter the if() block unless the first call to hasData() return true.

I'd recommend making that part of the test check whatever set of metrics it wants to be there and removing the hasData setting altogether.

labkey-jeckels (approved):

labkey-jeckels:
@labkey-sweta sorry, the change I just merged introduced a merge conflict. I think it should be straightforward to reconcile but let me know if you hit snags.

labkey-sweta:
> @labkey-sweta sorry, the change I just merged introduced a merge conflict. I think it should be straightforward to reconcile but let me know if you hit snags.
	
 Old code :    PEAK("Peak Area", true);
 New code :    TOTAL_PEAK("Total Peak Area (Precursor + Transition)");
The metric type drop shows Peak Area. Test is failing.

PR #436, opened 2021-10-05
Item 9533: App support for module defined sample assay results view configs to show in sample assays tabbed grid

cnathe:
#### Rationale
We want to allow for non-standard assays (i.e. those defined in customModules but not part of the assay designer infrastructure) to have a chance to participate in the Sample > Assays tabbed grid view within the LKB and LKSM apps. This PR adds a new SampleFileAssayResults query to be registered as the view to use in the app sample assays grid for showing Skyline Document related info for a run. It also updates a few places in the LKS grid/details view so that it uses the URLProvider override check to sent the user into the app UI when applicable for a sample link.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/649
* https://github.com/LabKey/platform/pull/2667
* https://github.com/LabKey/targetedms/pull/436
* https://github.com/LabKey/sampleManagement/pull/719
* https://github.com/LabKey/biologics/pull/1019
* https://github.com/LabKey/dataintegration/pull/129

#### Changes
* Add SampleFileAssayResults query and register it as the module's sample assay results view config to use in the apps
* SampleFileTable updates to opt in to checking for URLProvider overrides for the sample details URL
* SampleFileTable update to pull in sample type user defined fields to the default view when QueryForeignKey is established for the SampleName column


resources/queries/targetedms/SampleFileAssayResults.query.xml
```diff
@@ -0,0 +1,39 @@
...
+        <tables xmlns="http://labkey.org/data/xml">
+            <table tableName="SampleFileAssayResults" tableDbType="NOT_IN_DB">
+                <columns>
+                    <column columnName="Id">
+                        <isHidden>true</isHidden>
+                    </column>
+                    <column columnName="FileName">
+                        <url>/targetedms-showPrecursorList.view?id=${Id}</url>
+                    </column>
+                    <column columnName="PeptideGroupCount">
```
labkey-jeckels, line 12: PeptideGroupCount and MoleculeListCount are really two different names for the same underlying thing in the DB. We choose one based on the type of data that's in the files. Here, where we can't easily conditionalize, I think we're better off having a single column with title "Proteins/Molecule Lists"

resources/queries/targetedms/SampleFileAssayResults.sql
```diff
@@ -0,0 +1,14 @@
+SELECT
+SampleName,
+ReplicateId.RunId.FileName,
+ReplicateId.RunId.Id,
+AcquiredTime,
+CASE WHEN ReplicateId.RunId.SmallMoleculeCount = 0 THEN ReplicateId.RunId.PeptideGroupCount ELSE 0 END AS PeptideGroupCount,
```
labkey-jeckels: See comment from the .query.xml file

labkey-jeckels (approved):
Requests but no need for further review unless you have questions.

resources/queries/targetedms/SampleFileAssayResults.query.xml
cnathe: Got it. Done.

resources/queries/targetedms/SampleFileAssayResults.sql
cnathe: done

PR #437, opened 2021-10-14
Show multi-peptide/molecule, per-replicate plots 

labkey-jeckels:
#### Rationale
Users want to see all of the peptide or molecules that belong to a grouping at once so that they can compare their peaks against each other, and across replicates

#### Changes
* Introduce per-replicate, multi-peptide/molecule plots to the protein/molecule list details page
* Support selecting/unselecting from peptide/molecule list to narrow what's included in the plot
* Use SVG plots for chromatograms 
* Avoid showing "undefined" in QC summary panel when we don't know anything about the instrument that produced the data
* Steal Skyline code (and implement C# String hash code calculation) for molecule color selection
* Issue 43455: InternalError from org.labkey.targetedms.TargetedMSController.writeChart() when Linux fonts are missing

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -27,12 +27,12 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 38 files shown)


ankurjuneja (approved):

labkey-jeckels (commented):
The test looks good to me, @labkey-sweta. I can't officially approve because I opened the PR.

PR #453, opened 2021-12-06
Issue 44426: Slow queries for Skyline audit history on large databases

labkey-jeckels:
#### Rationale
We're seeing very poor perf on PanoramaWeb when viewing the Skyline audit log. It's because the DB isn't optimizing the CTE in the SQL very well.

#### Changes
* Remove AuditLogTraverse, AuditLog, AuditLogMessages, and AuditLogEvents from schema browser
* Refactor single document audit log view to use CTE scoped to just that document


ankurjuneja (approved):

labkey-jeckels:
Heard that we should target 21.11 instead. New PR coming momentarily.

PR #480, opened 2022-01-19
Issue 44557: Optimize queries within Panorama QC folders for faster plot rendering

labkey-jeckels:
#### Rationale
Fast pages are good. QC folders in Panorama are slower than we want, and there are some platform changes to help improve the SQL we run and how we process the data in Java.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2960

#### Changes
* Optimize SQL generation for QC metrics
* Pull redundant data (precursor and sample file info) out of the main metric query to improve SQL and Java perf (and memory footprint)
* Don't use reflection-based ResultSet->Java object mapping, instead pull directly from the ResultSet and use setters
* Misc minor code cleanup
* Cache annotation and QC metric configuration info for very short-term use

### DIFF
--- a/src/org/labkey/targetedms/model/SampleFileQCMetadata.java
+++ b/src/org/labkey/targetedms/model/SampleFileQCMetadata.java
@@ -1,22 +1,41 @@
 package org.labkey.targetedms.model;
 
+import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.targetedms.parser.SampleFile;
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
 public class SampleFileQCMetadata extends SampleFile
 {
-    boolean ignoreInQC;
     boolean inGuideSetTrainingRange;
+    private Set<Integer> _ignoredMetricIds = Collections.emptySet();
 
     @Nullable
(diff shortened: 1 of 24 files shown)


src/org/labkey/targetedms/model/SampleFileQCMetadata.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class SampleFileQCMetadata extends SampleFile
+{
+    boolean inGuideSetTrainingRange;
+    private Set<Integer> _ignoredMetricIds = Collections.emptySet();
+
+    @Nullable
+    public boolean isIgnoreInQC(int metricId)
+    {
+        return _ignoredMetricIds.contains(metricId) || _ignoredMetricIds.contains(-1);
```
ankurjuneja, line 21: Is this correct? I might be understanding incorrectly but presence of -1 indicates that metricId is not present in QCMetricExclusion.

ankurjuneja (approved):

src/org/labkey/targetedms/model/SampleFileQCMetadata.java
labkey-jeckels: I believe it's correct. I just pushed a commit with comments in both the Java code and SQL - can you see if it makes more sense now?

ankurjuneja: yeah it makes sense now with the context. Thanks.

PR #485, opened 2022-01-28
Issue 44765: Increase CacheFormatVersion to 15

vagisha:
#### Rationale
The current supported CacheFormatVersion in the targetedms module is 14. Skyline now supports version 15. The only new information added to version 15 is the "Import Time" which is the time the chromatograms were extracted. This can be ignored by Panorama so we can bump up the supported CacheFormatVersion to 15.


labkey-jeckels (changes requested):
It feels like we should update CachedFileHeaderStruct to match. Even if we don't read the new time info, shouldn't we update getStructSize() so that we don't grab other data from the wrong offsets?

vagisha:
> It feels like we should update CachedFileHeaderStruct to match. Even if we don't read the new time info, shouldn't we update getStructSize() so that we don't grab other data from the wrong offsets?

Done.

nickshulman (approved):
Looks good to me.

labkey-jeckels (approved):

PR #488, opened 2022-02-06
Issue 44820: Problems with library views and webparts

vagisha:
#### Rationale
- "Targeted MS Protein View" (and related Peptide / Precursor / Molecule / Molecule Precursor webparts) can be added in any targetedms folder type. However, the default behavior for these webparts is to display a view that is filtered to the current library targets. So, in non-library folders these webparts display an empty grid.
- Views meant for library folders (e.g. "Library Proteins", "All Folder Proteins" etc.) are visible in non-library folders.
- It is not possible to switch to the "default" view in these grids.
- In library folders, viewing a document with peptides displays a filtered grid with only those targets that are in the current library. The "All Precursors" view has to be selected to view all the targets in the document. Viewing a small molecule document however, displays all the targets in the document.

#### Changes
- In non-library folders the "Targeted MS Protein View" (an other) webparts display all the proteins (peptides etc.)
- "Library *" views are not be listed in the view chooser menu in non-library folders
- To get around not being able to switch to the "default" view in the "Targeted MS Protein View" (and other) webparts, display the library view without the view chooser or customize grid menus.  The "View All" button in the grid header, or the webpart title can be clicked to view the customizable grid. (based on patch from Josh)
- The nested grid for a single document displays all the targets in the document in any targetedms folder type. In library folders, a "Library Members" option is available in the views menu for viewing only those targets in the document that are in the current library in the folder.
- Removed unused views (FolderPoteins, FolderPeptides, FolderPrecursors, AllPrecursors).
- Added additional columns to the "LibraryProteins" view.
- Added a "RepresentivePrecursorCount" column to the PeptideGroup table. Use this as a filter in the "LibraryProteins" view instead of the "RepresentativeDataState" column so that the view can be used in both protein and peptide library folders.
- Removed 'libraryprecursor', 'librarymoleculeprecursor' and 'librarydocprecursor' queries. On PanoramaWeb there are only 3 custom view on these tables. Replaced their usage with 'precursor', 'moleculeprecursor' and 'experimentprecursor' queries respectively.
- Added an 'experimentmoleculeprecursor' query for viewing the nested grid for a single small molecule document. 
- 'experimentmoleculeprecursor' and 'experimentprecursor' have a LibraryMembers.qview.xml for viewing the targets from a single document that are in the current library in the folder.


labkey-jeckels (approved):

labkey-jeckels (approved):
Looks like a nice simplification. Thanks for the followup.

PR #490, opened 2022-02-15
Reduce JSON payload size in QC folder API, improve query caching

labkey-jeckels:
#### Rationale
We want QC folders to be faster

#### Changes
* Reduce JSON payload by not sending booleans that are false 99% of the time. Only send when true
* Pull more sample level metadata into the sample file section and out of the individual data rows
* Avoid round-tripping again for QC plot customizations that don't impact the data needed
* Reuse schema object to let us cache more tables and queries

### DIFF
--- a/webapp/TargetedMS/js/QCPlotHelperBase.js
+++ b/webapp/TargetedMS/js/QCPlotHelperBase.js
@@ -130,15 +130,18 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
 
         LABKEY.Ajax.request({
             url: LABKEY.ActionURL.buildURL('targetedms', 'GetQCPlotsData.api'),
-            success: this.processPlotData,
+            success: function(response) {
+                this.lastParsedResponse = JSON.parse(response.responseText);
+                this.processPlotData();
+            },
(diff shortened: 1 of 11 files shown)


webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -130,15 +130,18 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
...
             failure: LABKEY.Utils.getCallbackWrapper(this.failureHandler),
             scope: this,
             jsonData: plotsConfig
         });
     },
 
-    processPlotData: function(response) {
-        var parsed = JSON.parse(response.responseText);
+    processPlotData: function() {
+        var parsed = this.lastParsedResponse;
```
ankurjuneja, line 144: I can't remember the flow, but would this be always present in processPlotData? Would it make sense to check if (this.lastParsedResponse)?

ankurjuneja (approved):

webapp/TargetedMS/js/QCPlotHelperBase.js
labkey-jeckels: Good suggestion. I tested and if you click to change the plotting options before the AJAX call is complete it won't be initialized yet. I added a check

PR #495, opened 2022-02-23
Epic 9946: Ability to "hide" certain peptides/small molecules (AutoQC)

labkey-bpatel:
#### Rationale
Ability to hide certain peptide/small molecules in AutoQC plot page. This feature will help the scientist to filter the “good” peptides/molecule list, but still allow them to view the rest. 

#### Related Pull Requests
* https://github.com/LabKey/PanoramaPremiumModules/pull/77

#### Changes
* Custom queries and metadata for Peptide/Molecule precursors that groups precursors from different skyline documents
* Server side updates to only show Included precursors. 
* Add 'Show Excluded Precursors' checkbox to the plot panel.
* Schema update to add 'ExcludedPrecursors' table.


resources/queries/targetedms/QCGroupingPrecursorMolecules.sql
```diff
@@ -0,0 +1,40 @@
...
+
+SELECT
+    *,
+    (CASE
+         WHEN molPrec.precursorIdentifier IN (SELECT precursorIdentifier FROM targetedms.ExcludedPrecursors) THEN 'Excluded'
+         ELSE 'Included'
+        END) AS markedAs,
+FROM
+     (SELECT
+        ((CASE WHEN molPrec.MoleculeId.customIonName IS NOT NULL THEN molPrec.MoleculeId.customIonName
```
labkey-jeckels: Unfortunately, you can have any combination of null and non-null values for these two columns. We should include both if they're non-null

resources/queries/targetedms/QCGroupingPrecursorMolecules.sql
```diff
@@ -0,0 +1,40 @@
...
+        mol.peptideGroupId.Label,
+        -- mol.PeptideGroupId.RunId.File,
+        molPrec.MoleculeId.customIonName,
+        molPrec.MoleculeId.ionFormula,
+        molPrec.charge,
+        molPrec.mz,
+        molPrec.massMonoisotopic,
+        molPrec.massAverage
+        FROM
+            targetedms.Molecule mol
```
labkey-jeckels: This query will be more efficient if you just use the MoleculePrecursor table, and use the lookup syntax to get to its MoleculeId instead of doing an explicit join here. That way we won't have to separately filter both tables on Container

resources/queries/targetedms/QCGroupingPrecursorPeptides.sql
```diff
@@ -0,0 +1,34 @@
...
+        (pep.PeptideModifiedSequence || ',' || cast(prec.charge AS VARCHAR) || ',' || cast(round(prec.mz, 3) AS VARCHAR)) AS precursorIdentifier, -- "key field" required for 'requireSelection' to work
+        pep.PeptideModifiedSequence AS peptideSequence,
+        -- pep.PeptideGroupId.RunId.File,
+        pep.PeptideGroupId.Label,
+        prec.charge,
+        prec.mz,
+        prec.neutralMass,
+        prec.IsotopeLabelId.Name
+        FROM
+        targetedms.Peptide pep
```
labkey-jeckels: Same thing on this join - if you just use Precursor and get to the peptide columns via the lookup syntax it'll be faster.

resources/queries/targetedms/QCGroupingPrecursorPeptides.sql
```diff
@@ -0,0 +1,34 @@
...
+        LEFT JOIN
+        targetedms.Precursor prec
+        ON
+        pep.PeptideModifiedSequence = prec.PeptideId.modifiedPeptideDisplayColumn
```
labkey-jeckels: While this ON clause will go away based on my suggestion above, we should do joins on the Id columns and not the peptide sequence columns.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -312,18 +318,28 @@ private void retainIfPreferred(Map<String, QCMetricConfiguration> preferredConfi
...
     @NotNull
-    private Map<Long, RawMetricDataSet.PrecursorInfo> loadPrecursors(TargetedMSSchema schema) throws SQLException
+    private Map<Long, RawMetricDataSet.PrecursorInfo> loadPrecursors(TargetedMSSchema schema, Map<Long, Object> excludedPrecursorsIds, boolean showExcludedPrecursors) throws SQLException
     {
         Map<Long, RawMetricDataSet.PrecursorInfo> precursors = new HashMap<>();
 
         DecimalFormat format = new DecimalFormat();
         format.setMinimumFractionDigits(4);
 
+        Collection<Map<String, Object>> excludedPrecursors = new TableSelector(schema.getTable("ExcludedPrecursors")).getMapCollection();
```
labkey-jeckels: Please create a simple bean class instead of using a map.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -312,18 +318,28 @@ private void retainIfPreferred(Map<String, QCMetricConfiguration> preferredConfi
...
         try (ResultSet rs = new TableSelector(schema.getTable(TargetedMSSchema.TABLE_PRECURSOR)).getResultSet(false))
         {
             while (rs.next())
             {
+                if (!showExcludedPrecursors && isExcludedPrecursorPeptide(excludedPrecursors,
+                        rs.getString("ModifiedSequence"),
+                        rs.getInt("Charge"),
+                        rs.getDouble("MZ")))
+                {
+                    excludedPrecursorsIds.put(rs.getLong("Id"), null);
```
labkey-jeckels, line 342: Why is excludedPrecursorIds a map? If the value is always null, please use a Set.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -344,6 +371,24 @@ private void retainIfPreferred(Map<String, QCMetricConfiguration> preferredConfi
         return precursors;
     }
 
+    private boolean isExcludedPrecursorPeptide(Collection<Map<String, Object>> excludedPrecursors, String modifiedSeq, int charge, double mz)
+    {
+        String peptidePrecursorIdentifier = modifiedSeq + "," + charge + "," + DECIMAL_FORMAT.format(mz);
```
labkey-jeckels: This seems fragile since it needs to match up with how the SQL query is building the string. Can we find a way to consolidate implementations and/or add comments saying that they have to be kept in sync.

labkey-jeckels (changes requested):

resources/schemas/dbscripts/postgresql/targetedms-22.001-22.002.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE targetedms.ExcludedPrecursors
+(
+    rowId SERIAL NOT NULL,
+    precursorIdentifier VARCHAR(1000),
+    Container ENTITYID NOT NULL,
+
+    CONSTRAINT PK_EXCLUDEDPRECURSORS_ROWID PRIMARY KEY (rowId),
+    CONSTRAINT FK_EXCLUDEDPRECURSORS_CONTAINER FOREIGN KEY (Container) REFERENCES core.Containers(EntityId),
+    CONSTRAINT UQ_EXCLUDEDPRECURSORS_PRECURSOR_IDENTIFIER UNIQUE (precursorIdentifier)
```
labkey-jeckels: This should include Container, as you might exclude the same precursor in two different folders.

resources/queries/targetedms/QCGroupingPrecursorMolecules.sql
labkey-bpatel: done

resources/queries/targetedms/QCGroupingPrecursorMolecules.sql
labkey-bpatel: done

resources/queries/targetedms/QCGroupingPrecursorPeptides.sql
labkey-bpatel: That makes sense

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-bpatel: done

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-bpatel: Now re-implemented.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-bpatel: Using a bean class, so no longer using this approach

resources/queries/targetedms/QCGroupingPrecursorMolecules.sql
```diff
@@ -0,0 +1,42 @@
...
+        GROUP BY
+            molPrec.GeneralMoleculeId.PeptideGroupId.Label,
+            molPrec.IonFormula,
+            molPrec.customIonName,
+            molPrec.charge,
+            molPrec.mz,
+            molPrec.massAverage,
+            molPrec.massMonoisotopic) mp
+    LEFT JOIN
+    (SELECT *, 'Excluded' AS markedAs FROM targetedms.ExcludedPrecursors) exclPrec
```
labkey-jeckels:
```suggestion
    FROM targetedms.ExcludedPrecursors exclPrec
```

resources/queries/targetedms/QCGroupingPrecursorMolecules.sql
```diff
@@ -0,0 +1,42 @@
...
+SELECT
+    mp.Id,
+    mp.Label,
+    mp.customIonName,
+    mp.ionFormula,
+    mp.charge,
+    mp.mz,
+    mp.massMonoisotopic,
+    mp.massAverage,
+    (CASE WHEN exclPrec.markedAs = 'Excluded' THEN 'Excluded' ELSE 'Included' END) AS markedAs
```
labkey-jeckels:
```suggestion
    (CASE WHEN exclPrec.RowId IS NULL THEN 'Included' ELSE 'Excluded' END) AS markedAs
```

resources/queries/targetedms/QCGroupingPrecursorPeptides.sql
```diff
@@ -0,0 +1,37 @@
...
+
+SELECT
+precPep.Id,
+precPep.modifiedSequence,
+precPep.Label,
+precPep.charge,
+precPep.mz,
+precPep.neutralMass,
+precPep.name,
+(CASE WHEN exclPrec.markedAs = 'Excluded' THEN 'Excluded' ELSE 'Included' END) AS markedAs
```
labkey-jeckels: Same suggestion as the small molecule variant

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -344,6 +374,24 @@ private void retainIfPreferred(Map<String, QCMetricConfiguration> preferredConfi
...
+
+    private boolean isExcludedPrecursorMolecule(Collection<ExcludedPrecursor> excludedPrecursors, String customIonName,
+                                                String ionFormula, double massMonoisotopic, double massAverage, int charge, double mz)
+    {
+        return excludedPrecursors.stream().filter(ep -> Objects.equals(ep.getCustomIonName(), customIonName) &&
+                                                        Objects.equals(ep.getIonFormula(), ionFormula) &&
+                                                        Objects.equals(ep.getMassMonoisotopic(), massMonoisotopic) &&
+                                                        Objects.equals(ep.getMassAverage(), massAverage) &&
+                                                        Objects.equals(ep.getCharge(), charge) &&
+                                                        Objects.equals(ep.getMz(), mz)).count() == 1;
```
labkey-jeckels: Here you're looking for exact matches in the numeric values but in the SQL you're rounding to 4 decimal places. Should we be consistent? Can we always use the exact match?

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -344,6 +374,24 @@ private void retainIfPreferred(Map<String, QCMetricConfiguration> preferredConfi
         return precursors;
     }
 
+    private boolean isExcludedPrecursorPeptide(Collection<ExcludedPrecursor> excludedPrecursors, String modifiedSeq, int charge, double mz)
+    {
+        return excludedPrecursors.stream().filter(ep -> Objects.equals(ep.getModifiedSequence(), modifiedSeq) &&
```
labkey-jeckels: You should be able to use anyMatch() here instead of filter() which would be a tiny bit more efficient. Since the DB doesn't have a constraint to prevent duplicates anymore, this would also handle the case where someone managed to double-insert an exclusion

labkey-jeckels (approved):
A couple of small followup suggestions but no need for additional review unless you want it.

resources/queries/targetedms/QCGroupingPrecursorMolecules.sql
labkey-bpatel: done!
good suggestion, thanks!

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-bpatel: Now using exact match, removed rounding to 4 decimal places. I was trying something out since i wasnt getting expected results, then forgot to remove the rounding.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
labkey-bpatel: Good point! Now done

PR #498, opened 2022-03-03
Issue 44313: Dual-metric QC plots don't respect date filters when showing guide set data

ankurjuneja:
#### Rationale
This PR fixes the QC plots when when we're plotting two different metrics at the same time in case of displaying reference guideset with custom date range.

### DIFF
--- a/webapp/TargetedMS/js/QCPlotHelperBase.js
+++ b/webapp/TargetedMS/js/QCPlotHelperBase.js
@@ -196,30 +196,6 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
 
         }
 
-        if (this.showExpRunRange && this.filterPoints) {
-
-            for (let i = 0; i < plotDataRows.length; i++) {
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
@ankurjuneja I pulled down the FB and there must be an off-by-one error or something along those lines. I'm testing with SProCoPTutorial.zip from the sample data directory with a guide set that includes the first five samples (2013-08-09 through 2013-08-15 00:00:04).

Screen Shot 2022-03-03 at 10 15 44 AM

The plot is missing a point from the guide set range and it's connecting the last point to the first point with a line.

ankurjuneja:
@labkey-jeckels I think this PR is ready for re-review.

labkey-jeckels:
@ankurjuneja I'm seeing slightly different, but still incorrect, behavior. It's now connecting back to a different point.
image

labkey-jeckels (approved):
Looks good, thanks.

PR #499, opened 2022-03-05
Let admins save Panorama QC plot default configurations

labkey-jeckels:
#### Rationale
In preparing to let users submit QC folders to Panorama Public, we want to let admins share their QC plot configuration with other users

#### Changes
* New menu option for admins to save their current view as the default
* New menu option to let everyone revert to the default, whether it's the admin-saved default or the default default

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSModule.java
+++ b/src/org/labkey/targetedms/TargetedMSModule.java
@@ -17,6 +17,7 @@
 package org.labkey.targetedms;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.poi.ss.formula.functions.Na;
 import org.jetbrains.annotations.NotNull;
 import org.labkey.api.assay.sample.SampleAssayResultsConfig;
 import org.labkey.api.assay.sample.SampleAssayResultsService;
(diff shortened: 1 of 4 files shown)


src/org/labkey/targetedms/TargetedMSModule.java
```diff
@@ -507,6 +513,24 @@ protected void init()
...
 
+        Portal.registerNavTreeCustomizer("Targeted MS QC Plots", viewContext -> {
+            List<NavTree> result = new ArrayList<>();
+            if (viewContext.hasPermission(AdminPermission.class))
+            {
+                NavTree save = new NavTree("Save as Default View");
+                save.setScript("LABKEY.targetedms.PlotSettingsUtil.saveAsDefault()");
+                result.add(save);
+            }
+            if (!viewContext.getUser().isGuest())
```
labkey-bpatel, line 524: This means even the Readers would be able to revert to default, but RevertToDefaultQCPlotSettingsAction requires AdminPermission which would throw permission error.

labkey-jeckels: Thanks. I forgot to update the annotations on RevertToDefaultQCPlotSettingsAction after I copy/pasted the code to make the action. Fixed.

labkey-bpatel (approved):

PR #500, opened 2022-03-10
Issue 45015: peptide names inside QC plot legends are not displayed correctly

labkey-jeckels:
#### Rationale
When abbreviating peptide sequences, we shouldn't be including the mz part of the description as that's tacked on separately

#### Changes
* Trim the charge state and mz when creating unique abbreviations for each peptide sequence

### DIFF
--- a/webapp/TargetedMS/js/QCPlotLegendHelper.js
+++ b/webapp/TargetedMS/js/QCPlotLegendHelper.js
@@ -37,6 +37,11 @@ if (!LABKEY.targetedms.QCPlotLegendHelper) {
         addString: function (name, isSequence) {
             var dict = isSequence ? this.peptidePrefixDictionary : this.customIonPrefixDictionary;
 
+            // For peptides, only look at the sequence
+            if (isSequence) {
+                name = name.substring(0, name.indexOf(' '));
+            }
+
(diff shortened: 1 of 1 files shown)


webapp/TargetedMS/js/QCPlotLegendHelper.js
```diff
@@ -37,6 +37,11 @@ if (!LABKEY.targetedms.QCPlotLegendHelper) {
         addString: function (name, isSequence) {
             var dict = isSequence ? this.peptidePrefixDictionary : this.customIonPrefixDictionary;
 
+            // For peptides, only look at the sequence
+            if (isSequence) {
+                name = name.substring(0, name.indexOf(' '));
```
ankurjuneja, line 42: In this case, does name always have ' ' ? if name does not have ' ', then the substring will return ' '.

ankurjuneja (approved):

webapp/TargetedMS/js/QCPlotLegendHelper.js
labkey-jeckels: Indeed. Thanks. I thought I had tested this correctly via the JS console in the browser but my second argument to substring() was wrong.

PR #501, opened 2022-03-11
Light/heavy ratio support in Panorama reproducibility reports

labkey-jeckels:
#### Rationale
Users have requested that we support using the light/heavy precursors' ratio as the value of interest in Panorama's reproducibility report. This can be a more meaningful value than the raw intensity values of the two different precursors.

#### Changes
* Let users toggle between intensity and light/heavy ratio as the value of interest
* Show light/heavy chromatograms when displaying ratio
* Feed ratio data through the table and plots


webapp/passport/js/MxN/protein.js
```diff
@@ -72,11 +73,16 @@ protein =
                 '<div style="text-align: center"><a href="' + showPeptideUrl + 'id=' + protein.selectedPrecursor.PeptideId + '">Show peptide details</a></div></td></tr></table>';
         chromParent.append(childHtml);
 
-        $('a[name*=\'Chromatograms\'] > span').text('Chromatograms for ' + precursor.FullDescription);
+        const showRatio = $("input[name = 'intensityRatioToggle']:checked").val() === 'ratio';
+
+        $('a[name*=\'Chromatograms\'] > span').text('Chromatograms for ' + (showRatio ? precursor.Sequence : precursor.FullDescription));
```
ankurjuneja, line 78: encoding needed here?

webapp/passport/js/MxN/protein.js
```diff
@@ -697,5 +739,171 @@ protein =
...
+                geom: new LABKEY.vis.Geom.BarPlot({
+                    clickFn: function (event, row) {
+                        for (let i = 0; i < protein.precursors.length; i++) {
+                            if (protein.precursors[i].AbbreviatedLabel === row.sequence) {
+                                protein.selectPrecursor(protein.precursors[i].PrecursorId, true);
+                            }
+                        }
+                    },
+                    hoverFn: function (row) {
+                        return row.cvType + ': ' + row.cvMean.toFixed(1) + '%\n' +
```
ankurjuneja, line 780: encoding needed here?

webapp/passport/js/MxN/protein.js
```diff
@@ -697,5 +739,171 @@ protein =
...
+            geom: new LABKEY.vis.Geom.Point({
+                position: 'sequential',
+                color: 'timepoint',
+                shape: 'grouping',
+                opacity: 0.6,
+                size: 4
+            }),
+            aes: {
+                hoverText: function (row) {
+                    return format(row[valuePropertyName]) + '\n' +
```
ankurjuneja, line 839: encoding?

ankurjuneja (approved):

webapp/passport/js/MxN/protein.js
labkey-jeckels: Thanks for checking, but nope, text() handles the HTML encoding for us. https://api.jquery.com/text/

webapp/passport/js/MxN/protein.js
labkey-jeckels: Thanks for checking. The docs are less definitive here so I tested by intentionally injecting '<script>alert("test");</script>'. The script didn't execute and I saw that string in the tooltip itself.

PR #502, opened 2022-03-13
Issue 44507: Allow audit log entries to belong to multiple runs

labkey-martyp:
#### Rationale
If importing the same skyline file into multiple panorama projects on the same server, the audit log of the previous run will be overwritten by the audit log of subsequent runs.  This PR adds a junction table to allow mapping multiple runs to the same audit log entry.

#### Changes
* Add TargetedMS.RunAuditLogEntry junction table in database. Populate junction table from TargetedMS.AuditLogEntry.
* Remove VersionId from TargetedMS.AuditLogEntry
* Update logTree view to pull in RunId from junction table
* Handle multiple nodes with same EntryId in Audit log trees
* Update mutating operations on TargetedMS.AuditLogEntry VersionId column to instead update TargetedMS.RunAuditLogEntry.
* Update persisting and deleting operations on TargetedMS.AuditLogEntry to also update TargetedMS.RunAuditLogEntry


resources/schemas/dbscripts/postgresql/targetedms-22.002-22.003.sql
```diff
@@ -0,0 +1,35 @@
...
+CREATE INDEX IX_TARGETEDMS_RUNAUDITLOGENTRY_CONTAINER ON targetedms.RunAuditLogEntry (Container);
+CREATE UNIQUE INDEX UQ_TARGETEDMS_RUNAUDITLOGENTRY_RUN_AUDITLOGENTRY ON targetedms.RunAuditLogEntry(VersionId, AuditLogEntryId);
+
+-- Populate from existing AuditLogEntry and Run data
+INSERT INTO targetedms.RunAuditLogEntry (AuditLogEntryId, VersionId, Container, Created, CreatedBy, Modified, ModifiedBy)
+SELECT ale.EntryId AS AuditLogEntryId, ale.VersionId, r.Container, r.Created, r.CreatedBy, r.Modified, r.ModifiedBy FROM targetedms.AuditLogEntry ale                                                                                                                                  LEFT JOIN targetedms.Runs r ON ale.VersionId = r.id
+WHERE ale.VersionId IS NOT NULL;
+
+-- Drop this view
+SELECT core.fn_dropifexists('AuditLog', 'targetedms', 'VIEW', NULL);
```
labkey-jeckels: This won't do any harm, but all views should be deleted prior to running upgrade scripts (based on its inclusion in targeted ms-drop.sql)

resources/schemas/dbscripts/postgresql/targetedms-22.002-22.003.sql
```diff
@@ -0,0 +1,35 @@
+
+-- Junction table between Run and AuditLogEntry
+CREATE TABLE targetedms.RunAuditLogEntry
+(
+    RowId                   SERIAL NOT NULL,
+    VersionId               BIGINT NOT NULL,
+    AuditLogEntryId         INT NOT NULL,
+    Container               entityId NOT NULL,
```
labkey-jeckels: Can we avoid storing a Container here directly? We've only added it as a direct column to targetedms tables when it was strictly required for perf reasons. Everywhere else, we join to the run table (sometimes through a series of FKs) to figure out the Container. This makes it easier to move runs between containers, for example.

resources/schemas/dbscripts/postgresql/targetedms-create.sql
```diff
@@ -2,27 +2,24 @@ CREATE VIEW targetedms.AuditLog AS
...
-    SELECT entryId
-         , entryHash
-         , versionid AS RunId
-         , parentEntryHash
+    SELECT e.entryId
+         , e.entryHash
+         , e.parentEntryHash
     FROM targetedms.AuditLogEntry e
-    WHERE versionid IS NOT NULL
+    WHERE e.parentEntryHash = '(null)'
```
labkey-jeckels, line 9: Yuck. I didn't know we stored this as a string instead of just using null.

On PanoramaWeb.org we have an audit log table already. Since we don't have an index on parentEntryHash, I worry about perf here. I propose that we do an outer join to the new junction table here so that we can do a similar IS NOT NULL on VersionId.

src/org/labkey/targetedms/SkylineAuditLogManager.java
```diff
@@ -541,21 +560,21 @@ public void addAVersionTest() throws IOException, AuditLogException
             }
 
             assertNotNull(tree);
-            assertEquals(14, tree.getTreeSize());
+            assertEquals(15, tree.getTreeSize());
```
labkey-jeckels, line 563: This seems odd to me. Why are we getting additional entries in the tree? Perhaps we need to filter on the run ID to keep things consistent? I worry because places like SkylineAuditLogManager.deleteDocumentVersionLog() deletes the entries returned by buildLogTree() and we don't want deleting one document to mess up a separate copy of that document in a different folder.

src/org/labkey/targetedms/parser/skyaudit/AuditLogEntry.java
```diff
@@ -301,38 +303,42 @@ public boolean canBeHashed(AuditLogMessageExpander pExpander){
...
      */
-    public AuditLogEntry persist(){
+    public AuditLogEntry persist(User user, Container container){
         Table.insert(null, TargetedMSManager.getTableInfoSkylineAuditLogEntry(), this);
+
+        // save run audit log junction table entry
+        if (getVersionId() != null)
+        {
+            Map<String, Object> runAuditLogMap = new HashMap<>();
+            runAuditLogMap.put("VersionId", getVersionId());
```
labkey-jeckels: Can this call insertRunAuditLogEntry()?

labkey-jeckels (changes requested):
Thanks, this is looking good. Some questions and suggestions related to perf and data integrity.

src/org/labkey/targetedms/SkylineAuditLogManager.java
labkey-martyp: So the audit log tree now contains a node for every auditLogEntry/versionId combo.  This is necessary for adding the junction table, and it's exactly why SkylineAuditLogManager.deleteDocumentVersionLog does not delete auditLogEntries that belong to multiple runs.  This jest test contains many duplicate audit log entries across the runs. Only the leaf nodes of the audit log entry tree have the versionId so that is where new nodes are showing up in the tree.  That being said, I have somewhat limited testing data and scenarios here so happy to do some side by side testing on PanoramaWeb.org or wherever else there may be references.

resources/schemas/dbscripts/postgresql/targetedms-create.sql
labkey-martyp: You can't do an outer join in a CTE on PG.  I tried your suggestion before going to this one.  It is a head scratcher why we're using "(null)".  It's also used in CTEs building the audit log tree.  I can look at factoring that to use regular NULL but trying to not let this PR continue to grow.

resources/schemas/dbscripts/postgresql/targetedms-22.002-22.003.sql
labkey-martyp: Ok I removed container from this table and operations on the table

resources/schemas/dbscripts/postgresql/targetedms-22.002-22.003.sql
labkey-martyp: I missed that.  This has been removed

src/org/labkey/targetedms/parser/skyaudit/AuditLogEntry.java
labkey-martyp: Yes, refactored

labkey-jeckels (approved):
Perf looks good on PanoramaWeb-DR, so we're set to merge at your convenience.

PR #505, opened 2022-03-17
Issue 45050: Improve sample and replicate name handling in Panorama

labkey-jeckels:
#### Rationale
Skyline users are more used to seeing the replicate name displayed than the sample name. Additionally, we want to let them specify the canonical sample identifier to be used for joining with sample metadata.

#### Changes
* Look for an optional SampleIdentifier replicate annotation as the preferred name, falling back to SampleId and SampleName from the targetedms.SampleFile table.
* Show the replicate name more consistently in QC folders
* Small layout improvements to hover details
* Show more relevant columns in grids of Panorama samples/replicates


cnathe (commented):
I'll defer the final review for Ankur, but I did pull down this fb locally and verify that the Panorama / SM integration is still working as expected and is showing the expected grid columns based on the SampleFileAssayResults.sql changes here.

ankurjuneja (approved):

PR #506, opened 2022-03-23
Automation test for Add "light/heavy ratio" value type into reproducibility report

labkey-sweta:
#### Rationale
Automation test for Add "light/heavy ratio" value type into reproducibility report

#### Related Pull Requests
* 

#### Changes
1. Sample file : test/sampledata/TargetedMS/150604_VDBG_Hoofnagle_Precision_1119.sky.zip
2. Precursor webpart : test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
3. Test class : test/src/org/labkey/test/tests/targetedms/TargetedMSLightHeavyRatioTest.java


test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    public PrecursorsWebPart(WebDriver driver, int index)
+    {
+        super(driver, DEFAULT_TITLE, index);
+        waitForLoad();
+    }
+
+    @Override
+    public void clearCache()
```
labkey-jeckels: Is this override necessary? Seems like it could be deleted with no functional change.

test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
```diff
@@ -0,0 +1,168 @@
...
+            return getWrapper().isChecked(elementCache().ratio);
+    }
+
+    public PrecursorsWebPart selectValueType(String value)
+    {
+        getWrapper()._ext4Helper.selectComboBoxItem(elementCache().valueType, value);
+        return this;
+    }
+
+    public int getTableRowCount()
```
labkey-jeckels: Perhaps these intensity-related methods could be named getIntensityTableRowCount() and getIntensityTableElement() to better distinguish the pairs of methods

labkey-jeckels (approved):
A couple of suggestions but looking good.

test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
labkey-sweta: Done

test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
labkey-sweta: Deleted

test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
```diff
@@ -0,0 +1,162 @@
...
+        this(driver, 0);
+    }
+
+    public PrecursorsWebPart(WebDriver driver, int index)
+    {
+        super(driver, DEFAULT_TITLE, index);
+        waitForLoad();
+    }
+
+    public void waitForLoad()
```
labkey-danield, line 27: Should this be public?

test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
```diff
@@ -0,0 +1,162 @@
...
+    {
+        getWrapper().checkRadioButton(elementCache().intensity);
+        return this;
+    }
+
+    public PrecursorsWebPart selectRatio()
+    {
+        getWrapper().checkRadioButton(elementCache().ratio);
+        return this;
+    }
```
labkey-danield, line 63: I'm not very familiar with this feature, but is it possible to unselect intensity and ratio? If so would it make sense to pass in a boolean parameter to the selectIntensity and selectRatio methods that indicate if the option should be selected (parameter is true) or unselected (parameter is false)?

test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
```diff
@@ -0,0 +1,162 @@
...
+
+    public PrecursorsWebPart setSortBy(PrecursorsWebPart.SortBy option)
+    {
+        doAndWaitForElementToRefresh(() -> getWrapper().selectOptionByText(elementCache().sortBy, option.toString()),
+                getVisibleTable().append("//tbody/tr[1]/td[1]"), //Checking for staleness in the first element of the table
+                1000);
+        return this;
+    }
+
+    public Locator.XPathLocator getVisibleTable()
```
labkey-danield, line 40: Should this be public?

test/src/org/labkey/test/components/targetedms/PrecursorsWebPart.java
```diff
@@ -0,0 +1,162 @@
...
+    {
+        return elementCache().intensityTable.append("//tbody/tr").findElements(this).size();
+    }
+
+    public int getRatioTableRowCount()
+    {
+        return elementCache().ratioTable.append("//tbody/tr").findElements(this).size();
+    }
+
+    public String getIntensityTableElement(int row, int column)
```
labkey-danield, line 89: may want to be careful about using the term "Element". I have seen other test libraries where element refers to the actual web element. Maybe replace it with "Value"?

PR #508, opened 2022-03-28
Automation test for Issue 44507

labkey-sweta:
#### Rationale
Automation test Issue 44507: Importing the same Skyline document in multiple containers associates the audit log solely with the last import

1. Import into container A, verify you can see the log there.
2. Import into container B. Verify you can see the log in both A and B.
3. Delete from A. Verify you can see the log in B.
4. Celebrate!

#### Related Pull Requests
* 

#### Changes
New test case : TargetedMSAuditLogTest .testAuditLogInTwoDifferentFolders()

### DIFF
--- a/test/src/org/labkey/test/tests/targetedms/TargetedMSAuditLogTest.java
+++ b/test/src/org/labkey/test/tests/targetedms/TargetedMSAuditLogTest.java
@@ -49,7 +49,6 @@ public class TargetedMSAuditLogTest extends TargetedMSTest
                 auditLog.getDataAsText(0, "MessageText"));
         assertEquals("End message is incorrect", "Managed results",
                 auditLog.getDataAsText(8, "MessageText"));
-
     }
 
     /*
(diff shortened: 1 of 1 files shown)


test/src/org/labkey/test/tests/targetedms/TargetedMSAuditLogTest.java
```diff
@@ -57,4 +51,40 @@ public void testAuditLogImported()
                 auditLog.getDataAsText(8, "MessageText"));
 
     }
+
+    /*
+        Test coverage for Issue 44507: Importing the same Skyline document in multiple containers associates the audit log solely with the last import
```
labkey-jeckels, line 55: Is the order of test execution guaranteed, so that we know that testAuditLogImported() has already happened and will still be in the DB?

test/src/org/labkey/test/tests/targetedms/TargetedMSAuditLogTest.java
```diff
@@ -57,4 +51,40 @@ public void testAuditLogImported()
...
+    {
+        log("Creating folder and importing same SKY file");
+        goToProjectHome();
+        String FOLDER_2 = "Folder 2 with same audit logs";
+        setUpFolder(FOLDER_2, FolderType.QC);
+        importData(AuditTrail_FILE);
+
+        log("Verifying logs are imported correctly in " + FOLDER_2);
+        DataRegionTable auditLog = getAuditLogs(FOLDER_2);
+        assertEquals("Invalid number of audit logs", 9, auditLog.getDataRowCount());
```
labkey-jeckels: We need to verify that we can see the audit data in *both* folders when both copies are in the DB. This is part of step 2 in my test description in the issue.

labkey-jeckels (changes requested):

labkey-danield (approved):
If you address the concerns Josh raised I'm ok with this PR.

test/src/org/labkey/test/tests/targetedms/TargetedMSAuditLogTest.java
labkey-sweta: The AuditTrail_FILE file is imported in doInit() method, so yes the skyline file is available for both the tests irrespective of order.

test/src/org/labkey/test/tests/targetedms/TargetedMSAuditLogTest.java
labkey-sweta: Done

labkey-jeckels (approved):

PR #515, opened 2022-04-05
Epic 9975 : Adding support for copying QC Folders

labkey-bpatel:
#### Rationale
In the interest of publishing QC Folders to Panorama public, handle exporting and importing of QC properties such as: 
New custom metrics and trace metrics, Metric configurations and settings, Guide sets, Excluded samples, Excluded Precursors, Annotations, and Plot settings from default view.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/3229

#### Changes
* Framework to handle folder export and import
* Export Plot settings from default view to .properties file
* Export metrics, exclusions, guidesets, annotations to .tsv file only if data is present
* Handle exporting of "name" instead of Id
* Save Plot settings from .properties file
* Import metrics, exclusions, guidesets, annotations from .tsv to appropriate tables
* Resolve "name" into Id during import
* Skip inserting row if the data is already present


src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,478 @@
...
+                            {
+                                prop.put(name, plotSettings.get(name));
+                            }
+                        }
+                        prop.store(out, null);
+
+                    }
+                    catch (Exception e)
+                    {
+                        LOG.error("Error exporting 'default view QC Plot settings'", e);
```
labkey-jeckels: Are you expecting errors here? Rethrowing is probably better to make sure we ultimately learn of them and adjust as needed

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,478 @@
...
+                try (TSVGridWriter tsvWriter = new TSVGridWriter(factory))
+                {
+                    tsvWriter.setApplyFormats(false);
+                    tsvWriter.setColumnHeaderType(ColumnHeaderType.FieldKey);
+                    PrintWriter out = vf.getPrintWriter(fileName);
+                    tsvWriter.write(out);
+                }
+                catch (Exception e)
+                {
+                    LOG.error("Error writing results to " + fileName, e);
```
labkey-jeckels: Same here - recommend rethrowing

src/org/labkey/targetedms/folderImport/QCFolderImporter.java
```diff
@@ -0,0 +1,110 @@
...
+                       ctx.getLogger().info("Starting data import from " + qcSetting.getSettingsFileName() + " into targetedms." + qcSetting.getTableName());
+                       numRows = qcSetting.importSettingsFromFile(ctx, panoramaQCDir, schema, ti, qus, errors);
+                       ctx.getLogger().info("Finished importing " + numRows + " rows from " + qcSetting.getSettingsFileName() + " into targetedms." + qcSetting.getTableName());
+                   }
+               }
+               catch (IOException | DuplicateKeyException | BatchValidationException | QueryUpdateServiceException | SQLException e)
+               {
+                   if (qcSetting.getSettingsFileName().equalsIgnoreCase(QCFolderConstants.QC_PLOT_SETTINGS_PROPS_FILE_NAME))
+                   {
+                       LOG.error("Error importing QC Plot settings from " + QCFolderConstants.QC_PLOT_SETTINGS_PROPS_FILE_NAME + ": " + e.getMessage(), e);
```
labkey-jeckels: Like with the other errors, this means that nothing will get written to the job's log file and the import will be reported as successful.

labkey-jeckels (commented):
From a quick scan, this all looks reasonable. A couple of suggestions around exception handling. I can take a closer look when dev is complete.

src/org/labkey/targetedms/folderImport/QCFolderImporter.java
labkey-bpatel: good point, rethrowing exception now

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,489 @@
...
+public enum PanoramaQCSettings
+{
+    METRIC_CONFIG (TargetedMSSchema.TABLE_QC_METRIC_CONFIGURATION, QCFolderConstants.QC_METRIC_CONFIGURATION_FILE_NAME)
+            {
+                @Override
+                public void writeSettings(VirtualFile vf, Container container, User user) throws Exception
+                {
+                    TargetedMSSchema schema = new TargetedMSSchema(user, container);
+                    TableInfo ti = schema.getTable(getTableName());
+                    assert ti != null;
```
labkey-jeckels: I always find this style of assert useless, though I know we have it scattered throughout our code. On a dev machine with assertions turned on, it will fail with an assert on this line, with no message. On production servers it'll fail with a NPE on the next line. What's the advantage to asserting?

I'm guessing this was added because IntelliJ complained that `ti` might be null. @labkey-matthewb maybe we should have a UserSchema.getTableOrFail() method (or whatever we call it) that throws an IllegalArgumentException if the schema doesn't have the table. We can annotate that method as NotNull and have it fail with an error that includes the table name that failed to resolve.

src/org/labkey/targetedms/folderImport/QCFolderImporter.java
```diff
@@ -0,0 +1,106 @@
...
+                       ctx.getLogger().info("Starting data import from " + qcSetting.getSettingsFileName() + " into targetedms." + qcSetting.getTableName());
+                       numRows = qcSetting.importSettingsFromFile(ctx, panoramaQCDir, schema, ti, qus, errors);
+                       ctx.getLogger().info("Finished importing " + numRows + " rows from " + qcSetting.getSettingsFileName() + " into targetedms." + qcSetting.getTableName());
+                   }
+               }
+               catch (IOException | DuplicateKeyException | BatchValidationException | QueryUpdateServiceException | SQLException e)
+               {
+                   if (qcSetting.getSettingsFileName().equalsIgnoreCase(QCFolderConstants.QC_PLOT_SETTINGS_PROPS_FILE_NAME))
+                   {
+                       throw new Exception("Error importing QC Plot settings from " + QCFolderConstants.QC_PLOT_SETTINGS_PROPS_FILE_NAME + ": " + e.getMessage(), e);
```
labkey-jeckels: Throwing an ImportException instead of the base class is likely a bit friendlier

src/org/labkey/targetedms/folderImport/QCFolderConstants.java
```diff
@@ -0,0 +1,21 @@
+package org.labkey.targetedms.folderImport;
+
+import org.labkey.targetedms.TargetedMSSchema;
+
+public class QCFolderConstants
+{
+    protected static final String QC_FOLDER_DATA_TYPE = "Panorama QC Folder Settings";
+    protected static final String QC_FOLDER_DIR = "PanoramaQC";
+    protected static final String CATEGORY = "TargetedMSLeveyJenningsPlotOptions";
```
labkey-jeckels: Can we get rid of the separate constant in TargetedMSController now?

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,489 @@
...
+
+            logMsg += col + ": '" + row.get(col) + "'" + (count == row.size()-2 ? "" : ", ") ;
+            ++count;
+        }
+
+        assert ti != null;
+        if (new TableSelector(ti, row.keySet(), filter, null).getRowCount() > 0)
+        {
+            logMsg += "] values already exists. Skipping";
+            ctx.getLogger().warn(logMsg);
```
labkey-jeckels: This feels like info, not warning

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,489 @@
...
+                try (TSVGridWriter tsvWriter = new TSVGridWriter(factory))
+                {
+                    tsvWriter.setApplyFormats(false);
+                    tsvWriter.setColumnHeaderType(ColumnHeaderType.FieldKey);
+                    PrintWriter out = vf.getPrintWriter(fileName);
+                    tsvWriter.write(out);
+                }
+                catch (IOException e)
+                {
+                    throw new IOException("Error writing results to " + fileName, e);
```
labkey-jeckels: Can this be consolidated? Seems overkill to catch IOException, throw a new IOException, and then catch that all within 5 lines

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,489 @@
...
+                    SQLFragment sql = new SQLFragment("SELECT qcAnnotationType.Name AS QCAnnotationTypeId, qcAnnotation.Description, qcAnnotation.Date ")
+                            .append(" FROM qcAnnotation")
+                            .append(" INNER JOIN qcAnnotationType")
+                            .append(" ON qcAnnotation.QCAnnotationTypeId = qcAnnotationType.Id");
+
+                    Map<String, TableInfo> tableMap = new HashMap<>();
+                    tableMap.put("qcAnnotation", ti);
+                    tableMap.put("qcAnnotationType", schema.getTable(TargetedMSSchema.TABLE_QC_ANNOTATION_TYPE));
+
+                    ResultsFactory factory = ()-> QueryService.get().selectResults(schema, sql.getSQL(), tableMap, null, true, true);
```
labkey-jeckels: Why do we need to pass in the tableMap here? Can't QueryService resolve the tables itself from the schema?

labkey-jeckels (changes requested):

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -330,17 +310,37 @@ public String getSettingsFileName()
...
+    {
+        return getTableInfo(user, container, getTableName(), cf);
+    }
+
+    private @NotNull TableInfo getTableInfo(User user, Container container, String tableName, @Nullable ContainerFilter cf)
+    {
+        TargetedMSSchema schema = new TargetedMSSchema(user, container);
+        TableInfo ti = schema.getTable(tableName, cf);
+        if (null == ti)
+            throw new NotFoundException(schema.getSchemaName() + "." + getTableName() + " not found in '" + container.getName() + "'");
```
labkey-jeckels: container.getPath() is a bit more helpful

labkey-jeckels (approved):

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -53,8 +54,10 @@ public void exportSettings(VirtualFile vf, Container container, User user) throw
                     List<ColumnInfo> userEditableCols = ti.getColumns().stream().filter(ci -> ci.isUserEditable()).collect(Collectors.toList());
                     SimpleFilter filter = SimpleFilter.createContainerFilter(container); //only export the ones that are defined in current container (and not the ones from the root container)
 
-                    ResultsFactory factory = ()-> QueryService.get().select(ti, userEditableCols, filter, null);
+                    Results results = QueryService.get().select(ti, userEditableCols, filter, null);
+                    ResultsFactory factory = ()-> results;
                     exportSettingsToTSV(vf, factory, getSettingsFileName(), getTableName());
+                    results.close();
```
labkey-adam: This and the six other similar changes in this commit don't look right to me. The whole point of passing a factory into `exportSettingsToTSV()` (which then passes it into `TSVGridWriter`) is to allow the calling code to determine how a `Results` is generated, but to give the responsibility for actually generating and closing the `Results` to the consuming code (TSVGridWriter, in this case). Passing in a factory allows the consumer to use try-with-resources to handle the results (much preferred over passing around open Results objects). This change is not going to fix any "CachedResultSet not closed" errors... invoking the factory and closing that Results here just executes an extra query and throws away the results; it doesn't in any way affect the Results generated by the downstream code after it re-invokes the factory.

Now I see the problem... `exportSettingsToTSV()` invokes `factory.get().countAll()`... this call to the factory generates the CachedResultSet that's never closed. I'm sure you want to execute the query only once (current code is executing it three times), so generate and stash a Results, check countAll(), and pass a new factory `()->results` to TSVGridWriter. You have to close the results in the count == 0 case, so might as well generate the Results in a try-with-resources. This means it will be double closed in most cases, but CachedResultSet doesn't care.

labkey-adam (changes requested):

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
labkey-adam: This is difficult to explain in a commit comment... happy to discuss, just let me know

labkey-bpatel: This change did fix the "CachedResultSet not closed" errors: 
https://teamcity.labkey.org/viewType.html?buildTypeId=LabKey_223Release_Premium_ModulesSuites_PanoramaPremiumPostgres&branch_LabKey_223Release_Premium_ModulesSuites=copy_qc_folders&tab=buildTypeStatusDiv

Sounds good, lets sync on Monday! Thanks!

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,533 @@
...
+                    Results results = QueryService.get().selectResults(schema, sql.getSQL(), null, null, true, true);
+                    ResultsFactory factory = ()-> results;
+                    exportSettingsToTSV(vf, factory, getSettingsFileName(), getTableName());
+                    results.close();
+                }
+
+                @Override
+                public int importSettingsFromFile(FolderImportContext ctx, VirtualFile panoramaQCDir, @Nullable TargetedMSSchema schema, @Nullable TableInfo ti, @Nullable QueryUpdateService qus, @Nullable BatchValidationException errors) throws Exception
+                {
+                    DataLoader loader = new TabLoader(Readers.getReader(panoramaQCDir.getInputStream(getSettingsFileName())), true);
```
labkey-jeckels: This and other TabLoader instances need to be closed, ideally in a try-with-resources block.

labkey-jeckels (changes requested):

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
labkey-adam: We discussed today, but for the record: I missed that invoking the factory doesn't execute the query... so it makes sense that this change fixed the exception and doesn't actually execute the query multiple times (as I had claimed). I still think it would be cleaner for each caller to simply pass in a query-executing factory and for `exportSettingsToTSV()` to invoke the factory in a try-with-resources, check countAll(), and pass on a `()->results` factory if appropriate.

labkey-bpatel: Hi Adam - just seeing your followup comment, thanks for clarifying and insights. I've made some updates, please review. Thank you!

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
labkey-bpatel: Done! Thanks, Josh!

labkey-adam:
I was suggesting you pass in a factory to `exportSettingsToTSV()`. You're generating and passing in a `Results` instead... that's fine, actually, it's cleaner than my suggestion. But the code generating the Results should (almost) always close it in a try-with-resources. See inline comments.

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+    METRIC_CONFIG (TargetedMSSchema.TABLE_QC_METRIC_CONFIGURATION, QCFolderConstants.QC_METRIC_CONFIGURATION_FILE_NAME)
+            {
+                @Override
+                public void exportSettings(VirtualFile vf, Container container, User user) throws Exception
+                {
+                    TableInfo ti = getTableInfo(user, container, null);
+                    List<ColumnInfo> userEditableCols = ti.getColumns().stream().filter(ci -> ci.isUserEditable()).collect(Collectors.toList());
+                    SimpleFilter filter = SimpleFilter.createContainerFilter(container); //only export the ones that are defined in current container (and not the ones from the root container)
+
+                    Results results = QueryService.get().select(ti, userEditableCols, filter, null);
```
labkey-adam: try-with-resources

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+                {
+                    TargetedMSSchema schema = new TargetedMSSchema(user, container);
+                    TableInfo ti = getTableInfo(user, container, null);
+                    SQLFragment sql = new SQLFragment("SELECT qcMetricConfig.Name AS metric, qcEnabledMetrics.enabled, qcEnabledMetrics.lowerBound, qcEnabledMetrics.upperBound, qcEnabledMetrics.cusumLimit FROM ")
+                            .append(ti, "qcEnabledMetrics")
+                            .append(" INNER JOIN ")
+                            .append(TargetedMSManager.getTableInfoQCMetricConfiguration(), "qcMetricConfig")
+                            .append(" ON qcEnabledMetrics.metric = qcMetricConfig.Id");
+
+                    Results results = QueryService.get().selectResults(schema, sql.getSQL(), null, null, true, true);
```
labkey-adam: try-with-resources

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+    QC_ANNOTATION_TYPE (TargetedMSSchema.TABLE_QC_ANNOTATION_TYPE, QCFolderConstants.QC_ANNOTATION_TYPE)
+            {
+                @Override
+                public void exportSettings(VirtualFile vf, Container container, User user) throws Exception
+                {
+                    ContainerFilter cf = ContainerFilter.getContainerFilterByName(ContainerFilter.Type.CurrentPlusProjectAndShared.name(), container, user);
+                    TableInfo ti = getTableInfo(user, container, cf);
+                    List<ColumnInfo> userEditableCols = ti.getColumns().stream().filter(ci -> ci.isUserEditable()).collect(Collectors.toList());
+
+                    Results results = QueryService.get().select(ti, userEditableCols, null, null);
```
labkey-adam: try-with-resources

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+                    exportSettingsToTSV(vf, results, getSettingsFileName(), getTableName());
+                    results.close();
```
labkey-adam: Remove... always use try-with-resources instead of explicitly calling close() (unless there's a really good reason)

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+                    TableInfo ti = getTableInfo(user, container, null);
+                    SQLFragment sql = new SQLFragment("SELECT qcAnnotationType.Name AS QCAnnotationTypeId, qcAnnotation.Description, qcAnnotation.Date FROM ")
+                            .append(ti, "qcAnnotation")
+                            .append(" INNER JOIN ")
+                            .append(" (SELECT * FROM targetedms.qcAnnotationType ")
+                            .append(" UNION ")
+                            .append(" SELECT * FROM \"Shared\".targetedms.qcannotationtype) qcAnnotationType")
+                            .append(" ON qcAnnotation.QCAnnotationTypeId = qcAnnotationType.Id");
+
+                    Results results = QueryService.get().selectResults(schema, sql.getSQL(), null, null, true, true);
```
labkey-adam: try-with-resources

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+                            .append(ti, "replicateAnnotation")
+                            .append(" INNER JOIN ")
+                            .append(TargetedMSManager.getTableInfoReplicate(), "replicate")
+                            .append(" ON replicateAnnotation.replicateId = replicate.Id")
+                            .append(" INNER JOIN ")
+                            .append(TargetedMSManager.getTableInfoRuns(), "runs")
+                            .append(" ON replicate.RunId = runs.Id")
+                            .append(" WHERE Source != 'Skyline' ");
+
+                    Results results = QueryService.get().selectResults(schema, sql.getSQL(), null, null, true, true);
```
labkey-adam: try-with-resources

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+                            .append(TargetedMSManager.getTableInfoReplicate(), "replicate")
+                            .append(" ON qcMetricExclusion.ReplicateId = replicate.Id")
+                            .append(" LEFT JOIN ")
+                            .append(TargetedMSManager.getTableInfoQCMetricConfiguration(), "qcMetricConfig")
+                            .append(" ON qcMetricExclusion.MetricId = qcMetricConfig.Id")
+                            .append(" INNER JOIN ")
+                            .append(TargetedMSManager.getTableInfoRuns(), "runs")
+                            .append(" ON replicate.RunId = runs.Id");
+
+                    Results results = QueryService.get().selectResults(schema, sql.getSQL(), null, null, true, true);
```
labkey-adam: try-with-resources

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+            throw new NotFoundException(schema.getSchemaName() + "." + getTableName() + " not found in '" + container.getPath() + "'");
+        return ti;
+    }
+
+    public void exportSettings(VirtualFile vf, Container container, User user) throws Exception
+    {
+        TableInfo ti = getTableInfo(user, container, null);
+        List<ColumnInfo> userEditableCols = ti.getColumns().stream().filter(ci -> ci.isUserEditable()).collect(Collectors.toList());
+
+        Results results = QueryService.get().select(ti, userEditableCols, null, null);
```
labkey-adam: try-with-resources

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,524 @@
...
+        exportSettingsToTSV(vf, results, getSettingsFileName(), getTableName());
+    }
+
+    protected void exportSettingsToTSV(VirtualFile vf, Results results, String fileName, String tableName) throws Exception
+    {
+        //try-with-resources to ensure Results instance is closed regardless of whether try statement executes as expected or throws an exception
+        try (results)
```
labkey-adam: Remote this try since callers are generating the `Results` (and therefore responsible for closing it)

labkey-adam (changes requested):

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
labkey-bpatel: Good point! done

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,535 @@
...
+                {
+                    List<Map<String, Object>> tsvData = getTsvData(panoramaQCDir, getSettingsFileName());
+                    List<Map<String, Object>> dataWithoutDuplicates = new ArrayList<>();
+
+                        tsvData.forEach(row -> {
+                            String metricValue = (String) row.get("metric");
+                            Integer metricId = getRowIdFromName(metricValue, getSettingsFileName(), getTableName(), TargetedMSSchema.TABLE_QC_METRIC_CONFIGURATION, Set.of("Id"), new SimpleFilter(FieldKey.fromParts("Name"), metricValue), ctx.getUser(), ctx.getContainer(), null);
+                            row.put("metric", metricId);
+                            getDataWithoutDuplicates(ctx, ti, row, dataWithoutDuplicates);
+                        });
```
labkey-jeckels: This is indented an extra level

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,535 @@
...
+    {
+        try
+        {
+            if (results.countAll() > 0)
+
+                try (TSVGridWriter tsvWriter = new TSVGridWriter(()-> results))
+                {
+                    tsvWriter.setApplyFormats(false);
+                    tsvWriter.setColumnHeaderType(ColumnHeaderType.FieldKey);
+                    PrintWriter out = vf.getPrintWriter(fileName);
```
labkey-jeckels: The PrintWriter should be closed via a try-with-resources as well

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,535 @@
...
+                }
+                catch (DuplicateKeyException | BatchValidationException | QueryUpdateServiceException | SQLException e)
+                {
+                    throw new Exception("Data from " + settingsFileName + " did not get imported into targetedms." + tableName, e);
+                }
+                return insertedRows.size();
+            }
+            else
+            {
+                throw new NotFoundException("Query update service for targetedms." + tableName + " not found.");
```
labkey-jeckels: NotFoundException is intended for HTTP requests where the client sent a RowId or other identifier that the server couldn't find. IllegalArgumentException would be more appropriate here.

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,535 @@
...
+            if (qus != null)
+            {
+                List<Map<String,Object>> insertedRows;
+                try
+                {
+                    insertedRows = qus.insertRows(user, container, dataWithoutDuplicates, errors, null, null);
+                }
+                catch (DuplicateKeyException | BatchValidationException | QueryUpdateServiceException | SQLException e)
+                {
+                    throw new Exception("Data from " + settingsFileName + " did not get imported into targetedms." + tableName, e);
```
labkey-jeckels: ImportException is a more specific subclass that seems appropriate here.

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,535 @@
...
+        {
+            String msg = "Error resolving '" + name + "' to its corresponding Id. Id not found in lookup table targetedms." + lookupTableName + ". ";
+            if (null == targetTable)
+            {
+                msg += "Unable to save QC Properties from " + fileName + ".";
+            }
+            else
+                msg += "Unable to import data from " + fileName + " into targetedms." + targetTable;
+
+            throw new NotFoundException(msg);
```
labkey-jeckels: See comment below about NotFoundException vs IllegalArgumentException

src/org/labkey/targetedms/folderImport/PanoramaQCSettings.java
```diff
@@ -0,0 +1,535 @@
...
+                {
+                    tsvWriter.setApplyFormats(false);
+                    tsvWriter.setColumnHeaderType(ColumnHeaderType.FieldKey);
+                    PrintWriter out = vf.getPrintWriter(fileName);
+                    tsvWriter.write(out);
+                }
+        }
+        catch (IOException | SQLException e)
+        {
+            throw new Exception("Error getting results from " + tableName, e);
```
labkey-jeckels: See other comment about ImportException vs Exception. There are a few other examples

labkey-jeckels (changes requested):
A few final cleanup items

labkey-adam (approved):
Calls to `exportSettingsToTSV()` look good to me. Josh is reviewing everything else.

labkey-jeckels (approved):

PR #518, opened 2022-04-08
Item 10239: Panorama QC Plot improvements/cleanup (part 1)

cnathe:
#### Rationale
We want to make the QC plots as useful as possible in their default configuration, optimize them for readability, make their reportings actionable by the user, keep them performant, and more. This PR is an initial set of changes to move in that direction. Changes include: allow for point and legend hover to highlight a series in the plot when viewing the "Show All Series in a Single Plot" option, remove the option to show "Small" plot size, add a max tick mark setting, change excluded points to use a different plot icon instead of altering the opacity, remove the extra padding on the right margin to maximize the plot region. 

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/518
* https://github.com/LabKey/platform/pull/3237
* https://github.com/LabKey/PanoramaPremiumModules/pull/89

#### Changes
* remove test code references to plot side "Small"
* remove test check for guide set based plot shapes
* update QC plot options to remove plot size and rearrange plot types checkboxes
* update QC plot point and line size, add mouse over behavior to highlight series and legend items, add shapeRange so that excluded points use open circle, remove extra right margin


webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1314,7 +1240,18 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
         }
     },
 
-    plotPointHover : function(event, row, layerSel, point, valueName) {
+    legendMouseOver : function(data, item) {
+        if (data.name) {
+            // in the multi series case, the name as the series label appended, so use the hoverText instead
```
labkey-jeckels:
```suggestion
            // in the multi series case, the name has the series label appended, so use the hoverText instead
```

labkey-jeckels (approved):

cnathe:
One test failure on this fb: https://teamcity.labkey.org/viewLog.html?buildId=1774253&tab=buildResultsDiv&buildTypeId=LabKey_223Release_Community_ModuleSuites_TargettedMSPostgres&branch_LabKey_223Release_Community_ModuleSuites=panoramaQCPlotPart1

I just pushed an attempted fix. If that is successful, then I will get this merged.

PR #519, opened 2022-04-12
QC plot layout improvements: tooltip truncation, redundant titles, etc

labkey-jeckels:
#### Rationale
Issue 45204: Help users know the right AutoQC settings

We can improve the labels in Panorama QC plots and associated tooltips to reduce redundancy and increase accuracy

#### Changes
* Don't show plot subtitles that are redundant with web part headers, but do show them in exports
* Fix truncation on "Export to PNG" button
* Give users guidance on setting up AutoQC when they have an empty QC folder
* Don't lose the primary y-axis label when we're showing percent of mean, etc
* No need to show mz twice in the tooltip
* Improve formatting of values and layout in tulip


cnathe (approved):
Changes look good. I pulled down the fb locally and verified the QC plot related changes.

PR #523, opened 2022-04-19
Fix QC plot locators

labkey-tchad:
#### Rationale
QC plot IDs are not static. Need to update Locators.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3192

#### Changes
* Update Locator for `QCPlotsWebPart.getSVGPlotText`

### DIFF
--- a/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
@@ -286,7 +286,7 @@ public final class QCPlotsWebPart extends BodyWebPart<QCPlotsWebPart.Elements>
 
     public String getSVGPlotText(int plotPanelIndex, String plotIdSuffix)
     {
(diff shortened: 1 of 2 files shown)


test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
```diff
@@ -284,13 +284,20 @@ public List<QCPlot> getPlots()
         return plots;
     }
 
-    public String getSVGPlotText(String id)
+    public String getSVGPlotText(int plotPanelIndex, String plotIdSuffix)
```
labkey-matthewb, line 287: This function doesn't use plotPanelIndex?

labkey-matthewb (changes requested):

test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
labkey-tchad: Oops

labkey-matthewb (approved):

PR #531, opened 2022-05-10
Item 9548 - Show peak area or quality info in protein sequence view

ankurjuneja:
#### Rationale
This MR updates the settings passed to the protein view from the controller to add the heat map views of peptide's intensities and confidence scores.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3337

#### Changes
* new query to get peptide's intensities and confidence score


src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -4676,12 +4677,15 @@ public class ShowProteinAction extends SimpleViewAction<ChromatogramForm>
         @Override
         public ModelAndView getView(final ChromatogramForm form, BindException errors)
         {
+            String viewBy = getViewContext().getActionURL().getParameter("viewBy");
```
labkey-jeckels: Suggest adding this to the form object instead of pulling it separately from the URL. May want to introduce a ChromatogramForm subclass

src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -4836,14 +4840,17 @@ public static Integer addProteinSummaryViews(VBox box, PeptideGroup group, Targe
         if (group.getSequenceId() != null)
         {
             int seqId = group.getSequenceId().intValue();
-            List<String> peptideSequences = new ArrayList<>();
+            List<PeptideCharacteristic> peptideCharacteristics = new ArrayList<>();
             for (Peptide peptide : PeptideManager.getPeptidesForGroup(group.getId()))
             {
-                peptideSequences.add(peptide.getSequence());
+                var peptideCharacteristic = PeptideManager.getPeptideCharacteristic(peptide.getId());
+                peptideCharacteristic.setIntensityView(group.getViewBy() != null && group.getViewBy().equalsIgnoreCase("intensity"));
```
labkey-jeckels: Similar to my comment in the related PR, this doesn't feel like it should be a peptideCharacteristic level setting. They're all going to have the same value

src/org/labkey/targetedms/parser/PeptideGroup.java
```diff
@@ -46,6 +46,7 @@ public class PeptideGroup extends AnnotatedEntity<PeptideGroupAnnotation>
     private String _note;
 
     private String _altDescription;
+    private String _viewBy;
```
labkey-jeckels: This feels wrong - it's not something that we're persisting in the DB so it seems odd to add to the class that matches with the DB table

src/org/labkey/targetedms/query/PeptideManager.java
```diff
@@ -93,6 +94,24 @@ public static Collection<Peptide> getPeptidesForGroup(long peptideGroupId)
         return new SqlSelector(TargetedMSManager.getSchema(), sql).getArrayList(Peptide.class);
     }
 
+    public static PeptideCharacteristic getPeptideCharacteristic(long peptideId)
+    {
+        SQLFragment sql = new SQLFragment("SELECT X.Sequence, LOG10(MAX(X.Intensity)) AS Intensity, -ROUND(LOG10(MAX(X.Confidence)),4) AS Confidence FROM ");
```
labkey-jeckels: I think we want the raw qvalue for the confidence, not its log.

src/org/labkey/targetedms/query/PeptideManager.java
```diff
@@ -93,6 +94,24 @@ public static Collection<Peptide> getPeptidesForGroup(long peptideGroupId)
         return new SqlSelector(TargetedMSManager.getSchema(), sql).getArrayList(Peptide.class);
     }
 
+    public static PeptideCharacteristic getPeptideCharacteristic(long peptideId)
```
labkey-jeckels: We could easily get all of the peptides for a peptideGroup in a single query.

labkey-jeckels (changes requested):

src/org/labkey/targetedms/query/PeptideManager.java
ankurjuneja: I followed the specs, changing to raw value

src/org/labkey/targetedms/parser/PeptideGroup.java
ankurjuneja: yeah makes sense, I don't think it is needed anymore. removed.

src/org/labkey/targetedms/TargetedMSController.java
ankurjuneja: yep removed it.

src/org/labkey/targetedms/TargetedMSController.java
ankurjuneja: don't need it. removed it.

src/org/labkey/targetedms/query/PeptideManager.java
ankurjuneja: yep, done

src/org/labkey/targetedms/query/PeptideManager.java
labkey-jeckels: See comment in spec. OK to merge as-is and we can restore the log if I've misremembered.

labkey-jeckels (approved):

PR #537, opened 2022-05-16
Issue 45498: Show cross-linked peptide sequence info

labkey-jeckels:
#### Rationale
To better support multi attribute method analysis, Skyline added support for cross-linked peptides a couple of years ago. These are handled as special peptide modification annotations. Panorama can show the string representation but would greatly benefit from a more intuitive rendering, where we show all of the peptides involved in the linking and highlight the amino acids involved in the linking. We also want to improve the peptide mapping report for MAM folders to show more details about how and where the peptides are linked.

#### Changes
* Parse cross linked peptide sequences into a reusable data structure
* Render all peptides with link locations as modifications
* Unit tests for parsing, Selenium tests for peptide map rendering
* New/updated columns in Peptide Map report for proteins involved in linking and location of links
* More flexible column filtering for custom, MAM-related PTM report


src/org/labkey/targetedms/query/SamplePivotCustomizer.java
```diff
@@ -33,92 +31,157 @@
...
-            throw new IllegalArgumentException("Must have exactly one property named 'runIdField'");
+            throw new IllegalArgumentException("Must have exactly one property named '" + propertyName + "'");
         }
-        _runIdField = FieldKey.fromString(runIdFilter.iterator().next());
+        return null;
     }
 
-    @Override
-    public void customize(TableInfo tableInfo)
+    private Integer getRunId()
```
ankurjuneja: As runIds are of type long now, better to be consistent across.

src/org/labkey/targetedms/query/SamplePivotCustomizer.java
```diff
@@ -33,92 +31,157 @@
...
+                    clause.getParamVals().length == 1 &&
+                    clause.getParamVals()[0] != null)
             {
-                // Grab any filters that have been applied. Use QuerySettings, which will find the right values whether it's
-                // a GET or POST
-                QuerySettings settings = new QuerySettings(context, "query");
-                filterURL = settings.getSortFilterURL();
+                try
+                {
+                    // Get the sample names associated with that run
```
ankurjuneja: I am confused with the comment here.

ankurjuneja (approved):

src/org/labkey/targetedms/query/SamplePivotCustomizer.java
labkey-jeckels: We're a long way from 2 billion runs uploaded but agree best to parse as the right data type. Updated

src/org/labkey/targetedms/query/SamplePivotCustomizer.java
labkey-jeckels: Fixed. Agreed that it didn't make sense where it landed after the refactor

PR #549, opened 2022-06-02
Consolidate Panorama code into TargetedMS module

labkey-jeckels:
#### Rationale
Now that Panorama is fully a premium module, there's no need to maintain a split set of code

#### Changes
* Merge PanoramaPremium module into TargetedMS repo, maintaining history
* Reparent QCEmailNotifications into targetedms schema when present, create when not present
* Consolidate common classes


src/org/labkey/targetedms/TargetedMSSchema.java
```diff
@@ -1535,6 +1538,11 @@ public TableInfo getLookupTableInfo()
             return table;
         }
 
+        if(TABLE_QC_EMAIL_NOTIFICATIONS.equalsIgnoreCase(name))
```
ankurjuneja, line 1541:
```suggestion
        if (TABLE_QC_EMAIL_NOTIFICATIONS.equalsIgnoreCase(name))
```

ankurjuneja (approved):

PR #561, opened 2022-06-09
Test coverage for Issue 45440

labkey-sweta:
#### Rationale
Test coverage for Issue 45440: Moving range plot broken when choosing Percent of Mean or Standard Deviation y-axis options

#### Related Pull Requests
* 

#### Changes
Test case TargetedMSQCTest. testYAxisOptionsMovingRangePlot()

### DIFF
--- a/test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
+++ b/test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
@@ -878,26 +878,14 @@ public class TargetedMSQCTest extends TargetedMSTest
         String svgPlotText = qcPlotsWebPart.getSVGPlotText("tiledPlotPanel-2-precursorPlot0");
         assertFalse("Plot with standard deviations option is blank", svgPlotText.isEmpty());
         log("SVG content SD" + svgPlotText);
-        assertTrue("New plot is not as expected for standard deviations (y-axis) values ", svgPlotText.contains("-3\n" +
-                "-2\n" +
-                "-1\n" +
-                "0\n" +
-                "1\n" +
-                "2\n" +
-                "3\n" +
-                "4"));
+        assertTrue("New plot is not as expected for standard deviations (y-axis) values ", svgPlotText.contains("-3-2-101234"));
 
         log("Verifying percent of mean plots");
         qcPlotsWebPart.setScale(QCPlotsWebPart.Scale.PERCENT_OF_MEAN);
         svgPlotText = qcPlotsWebPart.getSVGPlotText("tiledPlotPanel-2-precursorPlot0");
         assertFalse("Plot with percent of mean option is blank", svgPlotText.isEmpty());
         log("SVG for POM " + svgPlotText);
-        assertTrue("New plot is not as expected for percent of mean (y-axis) values", svgPlotText.contains("90\n" +
-                "95\n" +
-                "100\n" +
-                "105\n" +
-                "110\n" +
-                "115"));
+        assertTrue("New plot is not as expected for percent of mean (y-axis) values", svgPlotText.contains("9095100105110115"));
     }
 
     private void verifyQCSummarySampleFileOutliers(String acquiredDate, String outlierInfo)
--- a/test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
+++ b/test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
@@ -877,6 +877,7 @@ public class TargetedMSQCTest extends TargetedMSTest
         qcPlotsWebPart.setScale(QCPlotsWebPart.Scale.STANDARD_DEVIATIONS);
         String svgPlotText = qcPlotsWebPart.getSVGPlotText("tiledPlotPanel-2-precursorPlot0");
         assertFalse("Plot with standard deviations option is blank", svgPlotText.isEmpty());
+        log("SVG content SD" + svgPlotText);
         assertTrue("New plot is not as expected for standard deviations (y-axis) values ", svgPlotText.contains("-3\n" +
                 "-2\n" +
                 "-1\n" +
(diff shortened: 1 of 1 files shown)


labkey-jeckels (changes requested):
If I'm reading the code correctly, this checks that the SVG changes when you toggle the y-scale option, but doesn't care what the SVG looks like. I believe this test would have passed before the bug fix.

We need look for something to verify that it's actually plotting the right thing. Maybe we could check for some of the numbers on the y-axis that change based on the y-scale option, and check the value shown in the tooltip for one of the data points.

Also, assertFalse() and assertTrue() are among the least informative possible ways to fail a test. At a minimum, including an error message when it fails or commenting on why it's doing that check would make troubleshooting any future failures much easier.

labkey-tchad (commented):
I'm not familiar enough with the feature or the bug to know whether this is a good way to check for regression. Doesn't seem like it verifies the plotted line is correct (mentioned in the issue).

labkey-jeckels (changes requested):
Conceptually the changes here look good (verifying numbers from the y-axis scale). I imagine you'll see the same failures locally as on TeamCity. The DOM/SVG is more complicated than having the numbers on separate lines.

You can also see this via the HTML artifact collected on the failure:

```
<g class="tick-text"><g><text x="65" y="221.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">-3</text></g><g><text x="65" y="201.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">-2</text></g><g><text x="65" y="182.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">-1</text></g><g><text x="65" y="163.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">0</text></g><g><text x="65" y="144.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">1</text></g><g><text x="65" y="125.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">2</text></g><g><text x="65" y="106.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">3</text></g><g><text x="65" y="86.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">4</text></g></g>
<g><text x="65" y="221.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">-3</text></g>
<text x="65" y="221.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">-3</text>
<g><text x="65" y="221.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">-3</text></g>
<g><text x="65" y="201.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">-2</text></g>
<text x="65" y="201.5" text-anchor="end" fill="#000000" style="font: 11px Roboto, arial, helvetica, sans-serif;">-2</text>
```

test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
```diff
@@ -860,6 +860,34 @@ public void testQCPlotExclusions()
...
+        QCPlotsWebPart qcPlotsWebPart = qcDashboard.getQcPlotsWebPart();
+        log("Enabling Moving range along with Levey-Jennings");
+        qcPlotsWebPart.checkPlotType(MovingRange, true);
+        qcPlotsWebPart.waitForPlots(PRECURSORS.length * 2, true);
+
+        log("Verifying standard deviations plots");
+        qcPlotsWebPart.setScale(QCPlotsWebPart.Scale.STANDARD_DEVIATIONS);
+        String svgPlotText = qcPlotsWebPart.getSVGPlotText("tiledPlotPanel-2-precursorPlot0");
+        assertFalse("Plot with standard deviations option is blank", svgPlotText.isEmpty());
+        log("SVG content SD" + svgPlotText);
```
labkey-jeckels: Do we always want to log this? Or should it be moved to only log when there's a failure?

test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
```diff
@@ -860,6 +860,34 @@ public void testQCPlotExclusions()
...
+        assertTrue("New plot is not as expected for standard deviations (y-axis) values ", svgPlotText.contains("-3-2-101234"));
+
+        log("Verifying percent of mean plots");
+        qcPlotsWebPart.setScale(QCPlotsWebPart.Scale.PERCENT_OF_MEAN);
+        svgPlotText = qcPlotsWebPart.getSVGPlotText("tiledPlotPanel-2-precursorPlot0");
+        assertFalse("Plot with percent of mean option is blank", svgPlotText.isEmpty());
+        log("SVG for POM " + svgPlotText);
+        assertTrue("New plot is not as expected for percent of mean (y-axis) values", svgPlotText.contains("9095100105110115"));
```
labkey-jeckels, line 888: This seems worthy of a comment that all of the values have been smashed together without whitespace. It's not very clear that this is checking for 90, 95, 100, 105, etc

labkey-jeckels (approved):

test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
labkey-sweta: Removed

test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
labkey-sweta: Done

PR #566, opened 2022-06-16
Issue 45701: Samples excluded for single metric don't get hidden in QC plots

labkey-jeckels:
#### Rationale
We should hide replicates hidden for individual and all metrics in QC plots unless you opt-in to seeing them. Right now we're only hiding the "all metric" exclusions.

#### Changes
* Filter the excluded data points before returning JSON to the client


vagisha (approved):
Where is this method called? Shouldn't that be updated as well?

labkey-jeckels:
> Where is this method called? Shouldn't that be updated as well?

@vagisha yes! I missed a file in my commit.

PR #571, opened 2022-06-25
Feature Request 45546: Offer to split graphs on precursor details page

labkey-jeckels:
#### Rationale
On our peptide details page we show a plot for all of the precursors of the peptide and then separate plots for each precursor. This is helpful because they may have wildly different intensities, so one can dwarf the other on the plot. 

Now we will do same on the precursor details page.

#### Changes
* New menu item on the chromatogram grid to toggle between split and combined views
* Hide menu item where it doesn't make sense 

Screen Shot 2022-06-24 at 6 18 27 PM


labkey-jeckels:
@brendanx67 any preferences on the default or naming for the button and menu items?

Previous behavior was that peptide/molecule details page showed split, and the precursor details page showed combined.

Revised behavior at the moment is that they both show combined by default.

ankurjuneja (approved):

PR #591, opened 2022-07-30
45833: Support new Skyline protein grouping XML

labkey-jeckels:
#### Rationale
Skyline is adding a new feature to assign peptides to not just a single protein, but a group of proteins when there are multiple possible matches for the observed peptides. This is captured via changes to the XML schema.

#### Changes
* Delete unused targetedms.protein table
* Move protein information from targetedms.peptidegroup into a new targetedms.protein table
* Parse updated XML with protein groups, retaining old code for parsing single-protein peptide groupings
* Migrate existing data
* Update reports and automated tests
* Add reporting on peptide group details page to show all proteins and let the user choose which one to show in the sequence coverage views


resources/queries/targetedms/Passport_proteins.sql
```diff
@@ -1,18 +1,18 @@
...
+  min(p.PeptideGroupId.id) AS proteinId @hidden
 
-FROM peptidegroup, runs
-WHERE runs.id = peptidegroup.runid
-      AND peptidegroup.sequenceid IS NOT NULL
-      and (select min(peptide.id) from peptide where peptide.peptidegroupid = peptidegroup.id and peptide.standardtype is not null) is null
-GROUP BY peptidegroup.sequenceid.seqid
\ No newline at end of file
+FROM protein p
+INNER JOIN peptide pep ON p.peptidegroupid = p.peptidegroupid AND pep.standardtype IS NOT NULL
```
ankurjuneja: p.peptidegroupid = p.peptidegroupid - this doesn't look right

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -1349,6 +1374,53 @@ else if (XmlUtil.isStartElement(reader, evtType, NOTE))
...
+            {
+                return result;
+            }
+            if (XmlUtil.isStartElement(reader, evtType, SEQUENCE))
+            {
+                result.setSequence(reader.getElementText().replaceAll("\\s+", ""));
+            }
+            else if (XmlUtil.isStartElement(reader, evtType, ANNOTATION))
+            {
+                // TODO?
```
ankurjuneja: is this for future?

src/org/labkey/targetedms/parser/SkylineDocumentParser.java
```diff
@@ -1349,6 +1374,53 @@ else if (XmlUtil.isStartElement(reader, evtType, NOTE))
...
+            }
+            else if (XmlUtil.isStartElement(reader, evtType, NOTE))
+            {
+                // TODO?
```
ankurjuneja: another todo

src/org/labkey/targetedms/view/precursorChromatogramsView.jsp
```diff
@@ -40,7 +40,7 @@
     </tr>
     <tr>
         <%
-            String fieldLabel = bean.getPeptideGroup().getSequenceId() != null ? "Protein" : "Group";
+            String fieldLabel = bean.getPeptideGroup().getProteins().isEmpty() ? "Group" : "Protein";
```
ankurjuneja: can "group" and "protein" be constants somewhere?

test/src/org/labkey/test/tests/targetedms/TargetedMSProteinGroupingTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
+import org.labkey.test.components.CustomizeView;
+import org.labkey.test.util.DataRegionTable;
+
+@Category({})
```
ankurjuneja, line 11: need to specify category here?

test/src/org/labkey/test/tests/targetedms/TargetedMSProteinGroupingTest.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    private void doInit()
+    {
+        setupFolder(FolderType.Experiment);
+        importData(SKY_FILE);
+    }
+
+    @Test
+    public void testProteinGrouping()
```
ankurjuneja, line 37: this test looks good, but I think there is more functionality that got added like the last bullet in Changes in this PR list.

ankurjuneja (changes requested):

test/src/org/labkey/test/tests/targetedms/TargetedMSProteinGroupingTest.java
labkey-sweta: Last week we decided to make the category blank for all targetedMS tests and control the category via teamcity composite build.

src/org/labkey/targetedms/view/precursorChromatogramsView.jsp
labkey-jeckels: I consolidated this snippet of code so it's not copied in three JSPs. I didn't see a lot of value in a constant since no other code is acting on the value being displayed (this isn't inherently linked with the name of the <protein> element in the XML files, for example, and the casing is different regardless.

resources/queries/targetedms/Passport_proteins.sql
labkey-jeckels: Thanks - one side should be pep.peptidegroupid. Fixed.

test/src/org/labkey/test/tests/targetedms/TargetedMSProteinGroupingTest.java
labkey-jeckels: Right - it would be good to check which sequence is being displayed and then click on a different protein in the grid and make sure that its sequence gets shown.

test/src/org/labkey/test/tests/targetedms/TargetedMSProteinGroupingTest.java
```diff
@@ -57,4 +71,11 @@ public void testProteinGrouping()
         peptideTable = new DataRegionTable.DataRegionFinder(getDriver()).withName("Peptides").waitFor();
         checker().verifyEquals("Incorrect group for the protein", group, peptideTable.getDataAsText(0, "PeptideGroupId"));
     }
-}
+
+    private String getCurrentSequence()
+    {
+        String unformattedString = (String) executeScript("return document.getElementById(\"peptideMap\").textContent;");
+        return unformattedString.replaceAll("[^A-Z]", "").trim().substring(1);
```
labkey-jeckels, line 82: This looks nice. But why the substring(1) - what extra character as you cutting away?

labkey-sweta: The textContent is getting the whole table text content along with this "The number displayed on each peptide box is the rank" text. 
When we replace everything but capital letters T stays back along with sequence so substring to get the sequence string.

ankurjuneja (approved):

test/src/org/labkey/test/tests/targetedms/TargetedMSProteinGroupingTest.java
labkey-jeckels: That definitely deserves a comment. We could also tweak the HTML we generate to make it easier for you to omit the caption - your call

labkey-sweta: Done..!

labkey-sweta: Added the comment.

PR #593, opened 2022-08-08
Automation test for Issue 46048

labkey-sweta:
#### Rationale
Automation test for Issue 46048: Configure Included and Excluded Precursors - blank or duplicate precursors in some scenarios

#### Related Pull Requests
https://github.com/LabKey/targetedms/pull/592

#### Changes
*

### DIFF
--- a/test/src/org/labkey/test/tests/panoramapremium/TargetedMSHidePeptidesAndMolecules.java
+++ b/test/src/org/labkey/test/tests/panoramapremium/TargetedMSHidePeptidesAndMolecules.java
@@ -17,6 +17,7 @@ import java.util.Arrays;
 public class TargetedMSHidePeptidesAndMolecules extends TargetedMSTest
 {
     protected static final String PeptidesOnlySkyFile = "QC_1.sky.zip";
+    protected static final String PeptidesOnlySkyFile_2 = "QC_2.sky.zip";
     protected static final String MoleculesOnlySkyFile = "SmMolLibA.sky.zip";
     protected static final String PeptidesAndMoleculesSkyFile = "smallmol_plus_peptides.sky.zip";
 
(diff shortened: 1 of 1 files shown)


test/src/org/labkey/test/tests/panoramapremium/TargetedMSHidePeptidesAndMolecules.java
```diff
@@ -78,6 +79,30 @@ public void testPeptideAndMoleculeExclusionAndInclusion()
...
+        Test coverage for : Issue 46048: Configure Included and Excluded Precursors - blank or duplicate precursors in some scenarios
+     */
+    @Test
+    public void testMultipleSkyFilesImport()
+    {
+        navigateToFolder(getProjectName(), PeptidesOnlySubfolder);
+        DataRegionTable table = gotoIncludeExcludeMenu();
+        checker().verifyEquals("Incorrect peptides with single import file", Arrays.asList("AGGSSEPVTGLADK", "VEATFGVDESANK"), table.getColumnDataAsText("modifiedSequence"));
+
+        log("Importing the second folder");
```
labkey-jeckels:
```suggestion
        log("Importing the second file folder");
```

test/src/org/labkey/test/tests/panoramapremium/TargetedMSHidePeptidesAndMolecules.java
```diff
@@ -78,6 +79,30 @@ public void testPeptideAndMoleculeExclusionAndInclusion()
         excludePeptideOrMolecule(MoleculesOnlySubFolder, 0, "Included");
     }
 
+    /*
+        Test coverage for : Issue 46048: Configure Included and Excluded Precursors - blank or duplicate precursors in some scenarios
```
labkey-jeckels, line 85: This is good and covers half of the behavior. It would be ideal to verify that you get a helpful error message when there is no data imported into the current folder. Maybe the test could create its own subfolder without importing any data to hit this case too.

labkey-jeckels (approved):
A couple of suggestions but no need for additional review

test/src/org/labkey/test/tests/panoramapremium/TargetedMSHidePeptidesAndMolecules.java
labkey-sweta: Done

PR #599, opened 2022-08-11
Issue 46081: uix_auditlogentry_document constraint violation

labkey-jeckels:
#### Rationale
Importing Skyline docs with overlapping and extending audit logs can sometimes cause a constraint violation due to non-deterministic ordering of the history during import

#### Changes
* Use a LinkedHashMap to guarantee ordering
* Add test coverage


labkey-martyp (approved):
Third time's a charm. Hope this gets the last of the corner cases.

PR #603, opened 2022-08-15
Update delete confirmation button text

labkey-tchad:
#### Rationale
Button text changed to "Yes, Delete" instead of "Confirm Delete"

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3589

#### Changes
* Update delete confirmation button text


labkey-susanh (changes requested):
I'm working on changes that will go back to using the more generic "confirmDelete.jsp" page since we need that for messaging about assay data linked from studies.  Tests should be happier once I've done that, so I think this PR can be closed.

PR #607, opened 2022-08-17
Fix chromatogram grid rendering

vagisha:
#### Rationale
Chromatograms grid in the peptide / molecule details page is displaying all the plots within a nested table in the first column making it appear like the table has extra headers.

#### Related Pull Requests
https://github.com/LabKey/targetedms/pull/598

#### Changes
Removed the nested table so that each plot is rendered in its own column


labkey-jeckels (approved):
I was seeing some other border/alignment issues on the peptide/molecule view when there were multiple precursors. Check out the new commit I pushed to completely omit those headers - it seems fine in my testing.

vagisha:
Looks good to me.  Thanks.

PR #614, opened 2022-08-31
Issue 46229: Use time-based ordering to jitter points in Panorama QC plots

labkey-jeckels:
#### Rationale
In Panorama QC folder plots, you can group data based on acquisition date. However, this randomly jitters the points within the day.  Like we do in the reproducibility plots, it would be better to order them based on their acquisition time. This would make it clearer when a change happens mid-day. 

#### Changes
* Configure to use sequential jittering, grouping by fragment


cnathe (approved):
let me know if you want me to pull down the FBs and do a manual testing pass as well

PR #622, opened 2022-10-10
Issue 46477: QCTrendPlotPanel switch to send guide set range elements to back instead of pulling all points to front

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46477

With the fix for [32277](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=32277), we fixed the issue in the Panorama QC Plots so that it moved the data points in the plot in front of the guide set range display boxes so that the hover / popup behavior would work as expected. As more data is added to a given folder, this runs into some performance issues moving that many data point elements in the DOM. With this change/fix, we flip that so instead we move the guide set range display boxes behind the data points in the DOM. This should have much better perf since there are far fewer of those elements in each plot.

#### Changes
* QCTrendPlotPanel switch to send guide set range elements to back instead of pulling all points to front

### DIFF
--- a/webapp/TargetedMS/js/QCTrendPlotPanel.js
+++ b/webapp/TargetedMS/js/QCTrendPlotPanel.js
@@ -1600,15 +1600,23 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
             }
         }
 
-        // Issue 32277: need to move the data points in front of the guide set range display
+        // Issue 46477: need to move the guide set range display behind the data points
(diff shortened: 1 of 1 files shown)


labkey-nicka (approved):

cnathe:
TeamCity runs against 22.7 look good:
https://teamcity.labkey.org/project.html?projectId=LabKey_227Release&tab=projectOverview&branch_LabKey_227Release=qcPlotPerf46477

PR #631, opened 2022-10-27
Avoid creating an empty PanoramaQC directory during folder import

labkey-bpatel:
#### Rationale
An empty PanoramaQC dir gets generated during a folder import via Pipeline.

Repro:
Import folder.xml via Pipeline (Ex. when importing folder.xml during for study import in ehr) - notice an empty 'PanoramaQC' folder under the pipeline directory.

#### Changes
* Update QCFolderImporter to import when 'PanoramaQC' dir is present in the archive and has data

### DIFF
--- a/src/org/labkey/targetedms/folderImport/QCFolderImporter.java
+++ b/src/org/labkey/targetedms/folderImport/QCFolderImporter.java
@@ -42,78 +42,61 @@ public class QCFolderImporter implements FolderImporter
     }
 
     @Override
-    public boolean isValidForImportArchive(FolderImportContext ctx) throws ImportException
-    {
-        //valid only if its a Targeted MS QC folder
-        boolean isTargetedMSFolder = false;
-        boolean isQCFolder = false;
-        if (null != ctx.getXml().getFolderType() && StringUtils.isNotBlank(ctx.getXml().getFolderType().getName()))
-        {
-            if (ctx.getXml().getFolderType().getName().equalsIgnoreCase("Targeted MS"))
-                isTargetedMSFolder = true;
-        }
-        if (isTargetedMSFolder)
-        {
-            ModulePropertyType[] moduleProps = ctx.getDocument().getFolder().getModuleProperties().getModulePropertyArray();
-            List<ModulePropertyType> targetedMSModuleProps = Arrays.stream(moduleProps).filter(modulePropertyType -> modulePropertyType.getPropertyName().equals("TargetedMS Folder Type")).collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


src/org/labkey/targetedms/folderImport/QCFolderImporter.java
```diff
@@ -38,55 +42,82 @@ public String getDescription()
...
-        TargetedMSSchema schema = new TargetedMSSchema(ctx.getUser(), ctx.getContainer());
+        return isQCFolder;
+    }
 
-        //iterate through PanoramaQCSettings enum values so that files get imported in that order/ordinal, since the lookup tables need to get populated first
-        for (PanoramaQCSettings qcSetting : PanoramaQCSettings.values())
+    @Override
+    public void process(@Nullable PipelineJob job, FolderImportContext ctx, VirtualFile root) throws Exception
+    {
+        if (isValidForImportArchive(ctx))
```
labkey-jeckels: This feels like the wrong approach to me - it will still create an empty directory if there wasn't any QC data in the folder archive.

src/org/labkey/targetedms/folderImport/QCFolderImporter.java
```diff
@@ -38,55 +42,82 @@ public String getDescription()
...
-           {
-               try
-               {
-                   long numRows;
+            if (null != job)
+            {
+                job.setStatus("IMPORT " + getDescription());
+            }
+
+            VirtualFile panoramaQCDir = root.getDir(QCFolderConstants.QC_FOLDER_DIR);
```
labkey-jeckels, line 50: Is this the line that ends up creating the directory? I'd say we should add a check to see if the info exists in the folder archive or otherwise not create it when it's not there

labkey-jeckels (changes requested):

src/org/labkey/targetedms/folderImport/QCFolderImporter.java
labkey-adam: Yes, it looks like `getDir()` will create the underlying directory if it doesn't exist

labkey-jeckels: @labkey-bpatel can you revisit this PR and close or update as appropriate?

labkey-bpatel: oh this totally fell through the cracks, revisiting now

labkey-jeckels (approved):

PR #632, opened 2022-11-01
Issue 46626: Show last 180 days in QC folders by default

labkey-jeckels:
#### Rationale
To help focus on the most relevant data and help with some perf concerns, we want to change the default date range to the last 180 days of data

#### Changes
* Default to last 180 days
* For Last N Days ranges, consider them relative to the most recent sample in the folder, not today's date

### DIFF
--- a/src/org/labkey/targetedms/query/ReplicateManager.java
+++ b/src/org/labkey/targetedms/query/ReplicateManager.java
@@ -53,21 +53,6 @@ public class ReplicateManager
             .getObject(sampleFileId, SampleFile.class);
     }
 
-    public static SampleFile getNewestSampleFile(Container container)
-    {
-        SQLFragment sqlFragment = new SQLFragment("SELECT s.* FROM ");
-        sqlFragment.append(TargetedMSManager.getTableInfoSampleFile(), "s");
-        sqlFragment.append(" INNER JOIN ").append(TargetedMSManager.getTableInfoReplicate(), "rep");
-        sqlFragment.append(" ON s.replicateId = rep.Id");
-        sqlFragment.append(" INNER JOIN ").append(TargetedMSManager.getTableInfoRuns(), "r");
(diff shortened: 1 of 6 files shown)


webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -179,7 +179,7 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
 
     calculateStartDateByOffset : function() {
         if (this.dateRangeOffset > 0) {
-            var todayMinusOffset = new Date();
+            var todayMinusOffset = this.maxAcquiredTime ? new Date(this.maxAcquiredTime) : new Date();
```
cnathe: should probably rename this variable now: `startDateByOffset`?

src/org/labkey/targetedms/query/ReplicateManager.java
```diff
@@ -52,6 +53,21 @@ public static SampleFile getSampleFile(long sampleFileId)
             .getObject(sampleFileId, SampleFile.class);
     }
 
+    public static SampleFile getNewestSampleFile(Container container)
```
cnathe: where is this called? I am not seeing it used in this PR.

cnathe (changes requested):
A couple of comments/questions. I will approve the PR after it is retargeted to 22.11 (once that is created today).

src/org/labkey/targetedms/query/ReplicateManager.java
labkey-jeckels: I removed the caller when I realized that the client already had the first/last dates available from an existing API call. I deleted this method.

webapp/TargetedMS/js/QCTrendPlotPanel.js
labkey-jeckels: Renamed

cnathe (approved):

PR #634, opened 2022-11-02
Streams returned from Files.walk() should be closed

labkey-adam:
#### Rationale
This one is probably closed properly, but let's be sure


labkey-jeckels (approved):
This refactor is fine, but yes, it's being closed. It caused real problems when it wasn't. https://github.com/LabKey/targetedms/pull/541

PR #636, opened 2022-11-03
22 11 fb bilinear regression fit

nickshulman:
Need to be fixed before merging: the calibration curve graph always draws as a straight line, even though the line for "Bilinear"  is supposed to be two straight lines with a turn in it, and "Linear in log space" produces a line which is curved when plotted on a linear graph.
image

#### Rationale
Issue 43922: Add support for "Linear in Log Space" and "Bilinear" regression fit options

#### Changes
Add support for "Bilinear" and "Linear in Log Space" regression fit values.


nickshulman:
Branch was misnamed so closing this and opening PR ##637 instead.

PR #649, opened 2022-12-21
Automation test for Issue 45701

labkey-sweta:
#### Rationale
Issue 45701: Samples excluded for single metric don't get hidden in QC plots Automation test

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45701

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels (changes requested):
The coverage that's here is all useful but doesn't cover the scenario described in the bug. Before setting the config to show excluded points line 943 we should confirm that the excluded point is NOT shown.

labkey-jeckels (approved):

PR #651, opened 2023-01-02
 FR 44763 Access mean and CV of any metric cross last N runs

ankurjuneja:
#### Rationale
This PR adds two new plots to Panorama QC - Access mean and CV for last N runs that would help analytical lab scientists to see trailing mean and CV values across last N runs for all the metrics so that the outliers will have less impact on the plots and they can get a better overall idea of the dataset. 

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3971

#### Changes
* new plots and their helpers

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -1279,7 +1279,7 @@ public class TargetedMSController extends SpringActionController
             // always query for the full range
             List<RawMetricDataSet> rawMetricDataSets = generator.getRawMetricDataSets(schema, qcMetricConfigurations, qcFolderStartDate, qcFolderEndDate, form.getSelectedAnnotations(), form.isShowExcluded(), form.isShowExcludedPrecursors());
             Map<GuideSetKey, GuideSetStats> stats;
-            if (form.includeTrailingCVPlot || form.includeTrailingMeanPlot)
+            if ((form.includeTrailingCVPlot || form.includeTrailingMeanPlot) && form.trailingRuns > 2)
             {
                 stats = generator.getAllProcessedMetricGuideSets(rawMetricDataSets, guideSets.stream().collect(Collectors.toMap(GuideSet::getRowId, Function.identity())), form.trailingRuns);
             }
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -1408,7 +1408,7 @@ public class TargetedMSController extends SpringActionController
             Map<GuideSetKey, GuideSetStats> stats;
             if ((form.includeTrailingCVPlot || form.includeTrailingMeanPlot) && form._trailingRuns > 2)
             {
-                stats = generator.getAllProcessedMetricGuideSets(rawMetricDataSets, guideSets.stream().collect(Collectors.toMap(GuideSet::getRowId, Function.identity())), form._trailingRuns, form._startDate, form._endDate);
+                stats = generator.getAllProcessedMetricGuideSets(rawMetricDataSets, guideSets.stream().collect(Collectors.toMap(GuideSet::getRowId, Function.identity())), form._trailingRuns);
             }
             else
             {
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -1281,7 +1281,7 @@ public class TargetedMSController extends SpringActionController
             Map<GuideSetKey, GuideSetStats> stats;
             if ((form.includeTrailingCVPlot || form.includeTrailingMeanPlot) && form.trailingRuns > 2)
             {
-                stats = generator.getAllProcessedMetricGuideSets(rawMetricDataSets, guideSets.stream().collect(Collectors.toMap(GuideSet::getRowId, Function.identity())), form.trailingRuns);
+                stats = generator.getAllProcessedMetricGuideSets(rawMetricDataSets, guideSets.stream().collect(Collectors.toMap(GuideSet::getRowId, Function.identity())), form.trailingRuns, form._startDate, form._endDate);
             }
             else
             {
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -18,6 +18,7 @@ package org.labkey.targetedms;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.keypoint.PngEncoder;
+import lombok.Builder;
 import lombok.Getter;
 import lombok.Setter;
 import org.apache.batik.dom.GenericDOMImplementation;
@@ -1191,7 +1192,7 @@ public class TargetedMSController extends SpringActionController
         @Getter @Setter  private boolean showExcluded;
         @Getter @Setter private boolean showReferenceGS;
         @Getter @Setter private boolean showExcludedPrecursors;
-        @Getter @Setter private int trailingRuns;
+        @Getter @Setter @Builder.Default private int trailingRuns = 10;
         @Getter @Setter private boolean includeTrailingMeanPlot;
         @Getter @Setter private boolean includeTrailingCVPlot;
 
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -1192,7 +1192,7 @@ public class TargetedMSController extends SpringActionController
         @Getter @Setter  private boolean showExcluded;
         @Getter @Setter private boolean showReferenceGS;
         @Getter @Setter private boolean showExcludedPrecursors;
-        @Getter @Setter @Builder.Default private int trailingRuns = 10;
+        @Getter @Setter private int trailingRuns = 10;
         @Getter @Setter private boolean includeTrailingMeanPlot;
         @Getter @Setter private boolean includeTrailingCVPlot;
 
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -18,6 +18,8 @@ package org.labkey.targetedms;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.keypoint.PngEncoder;
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.batik.dom.GenericDOMImplementation;
 import org.apache.batik.svggen.SVGGeneratorContext;
 import org.apache.batik.svggen.SVGGraphics2D;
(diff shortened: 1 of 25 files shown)


webapp/TargetedMS/js/TrailingCVPlotHelper.js
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (c) 2016-2018 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0: http://www.apache.org/licenses/LICENSE-2.0
+ */
+Ext4.define("LABKEY.targetedms.TrailingCVPlotHelper", {
+    extend: 'LABKEY.targetedms.QCPlotHelperBase',
+    statics: {
+        tooltips: {
+            'Trailing CV' : 'Blah '
```
ankurjuneja: @wendyinnis to provide.

webapp/TargetedMS/js/TrailingMeanPlotHelper.js
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (c) 2016-2018 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0: http://www.apache.org/licenses/LICENSE-2.0
+ */
+Ext4.define("LABKEY.targetedms.TrailingMeanPlotHelper", {
+    extend: 'LABKEY.targetedms.QCPlotHelperBase',
+    statics: {
+        tooltips: {
+            'Trailing Mean' : 'Blah '
```
ankurjuneja: @wendyinnis to provide

src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -1175,67 +1180,20 @@ public Object execute(QCMetricOutliersForm form, BindException errors)
...
-        private boolean _includeLJ;
-        private boolean _includeMR;
-        private boolean _includeMeanCusum;
-        private boolean _includeVariableCusum;
+        @Getter @Setter private int metricId;
+        @Getter @Setter private boolean includeLJ;
+        @Getter @Setter private boolean includeMR;
+        @Getter @Setter private boolean includeMeanCusum;
+        @Getter @Setter private boolean includeVariableCusum;
         private Date _startDate;
```
labkey-jeckels: Is there a reason you didn't convert `_startDate` and `endDate` too?

src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -1175,67 +1180,20 @@ public Object execute(QCMetricOutliersForm form, BindException errors)
...
-
-        public void setIncludeVariableCusum(boolean includeVariableCusum)
-        {
-            this._includeVariableCusum = includeVariableCusum;
-        }
+        @Getter @Setter private List<OutlierGenerator.AnnotationGroup> selectedAnnotations = Collections.emptyList();
+        @Getter @Setter  private boolean showExcluded;
+        @Getter @Setter private boolean showReferenceGS;
+        @Getter @Setter private boolean showExcludedPrecursors;
+        @Getter @Setter private int trailingRuns;
```
labkey-jeckels: Should we default this to a reasonable value like 10 so that if the API call doesn't include one, we don't pass 0 through to the calculations?

webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -602,6 +610,12 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
         if (plotType !== LABKEY.vis.TrendingLinePlotType.MovingRange && plotType !== LABKEY.vis.TrendingLinePlotType.LeveyJennings) {
             yScaleLabel = 'Sum of Deviations'
         }
+        if (plotType === LABKEY.vis.TrendingLinePlotType.TrailingMean) {
```
labkey-jeckels, line 613: Is this check needed? I think the others are just when we're overriding the label based on the data transformation. Not a big deal either way though

webapp/TargetedMS/js/QCPlotHelperWrapper.js
```diff
@@ -7,70 +7,151 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperWrapper", {
...
-    {
+    deepCloneArray: function (arr) {
+        return arr.map(function(item) {
+            // Check if the item is an array
+            if (Array.isArray(item)) {
+                // If it is, call deepCloneArray on it
+                return this.deepCloneArray(item);
+            } else if (typeof item === 'object' && item !== null) {
+                // If it's an object, use Object.assign to create a new object with the same properties
+                return Object.assign({}, item);
```
labkey-jeckels: I would have expected this to call deepCloneObject(). Otherwise it won't do a deep clone - it'll share the same objects as values, right?

webapp/TargetedMS/js/QCPlotHelperWrapper.js
```diff
@@ -7,70 +7,151 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperWrapper", {
...
+                if (new Date(data.fullDate).getTime() < firstGuideSetStartDate.getTime()) {
+                    precursors.data.splice(ind, 1);
+                    ind--;
+                }
+            }
+        }
+        else {
+            for (let ind = 0 ; ind < precursors.data.length; ind++) {
+                let data = precursors.data[ind];
+                if (!data.TrailingMean) {
```
labkey-jeckels: Is the assumption here that it'll either have both TrailingMean and TrailingCV or neither?

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -297,59 +302,84 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
+                fieldLabel: 'Tailing Last',
+                labelWidth: 65,
+                width: 115,
+                enableKeyEvents: true,
+                id : 'trailingRuns',
+                value: this.trailingRuns ? this.trailingRuns : this.runs > 10 ? 10 : this.runs,
+                hidden: true,
+                activeError: '',
+                allowDecimals: false,
+                minValue: 0,
```
labkey-jeckels: Shouldn't we set the lower bound to 2? 0 or 1 don't seem useful

webapp/TargetedMS/js/TrailingCVPlotHelper.js
```diff
@@ -0,0 +1,95 @@
...
+        else {
+            plotProperties['TrailingCV'] = 'TrailingCV';
+            let min = Math.min(...precursorInfo.data.map(function(object) {
+                return object.TrailingCV;
+            }));
+            let max = Math.max(...precursorInfo.data.map(function(object) {
+                return object.TrailingCV;
+            }));
+
+            // the number 20 is specified in the specs
```
labkey-jeckels:
```suggestion
            // Since 20 is a common limit for acceptable %CV, use that as the default range
```

webapp/TargetedMS/js/TrailingCVPlotHelper.js
```diff
@@ -0,0 +1,95 @@
...
+            if (min < 20 && max < 20) {
+                min = 0;
+                max = 20;
+            }
+
+            if (min > 20 || max > 20) {
```
labkey-jeckels, line 52:
```suggestion
            // expand range if we have values bigger than 20
            if (min > 20 || max > 20) {
```

src/org/labkey/targetedms/TargetedMSController.java
ankurjuneja: cause setter is using java.sql.Date and getter is java.util.Date.

webapp/TargetedMS/js/QCPlotHelperBase.js
ankurjuneja: I think it is needed, as in the next section it is getting a different conversion label if not set.

webapp/TargetedMS/js/QCPlotHelperWrapper.js
ankurjuneja: yep

webapp/TargetedMS/js/QCPlotHelperWrapper.js
ankurjuneja: correct

webapp/TargetedMS/js/QCTrendPlotPanel.js
ankurjuneja: yes

webapp/TargetedMS/js/TrailingCVPlotHelper.js
labkey-jeckels: Still need to swap this with real text. @wendyinnis

src/org/labkey/targetedms/model/QCPlotFragment.java
```diff
@@ -138,6 +138,19 @@ public JSONObject toJSON(boolean includeLJ, boolean includeMR, boolean includeMe
...
+                    dataJsonObject.put("TrailingMean", plotData.getTrailingMean());
+                }
+                if (includeTrailingCV)
+                {
+                    dataJsonObject.put("TrailingCV", plotData.getTrailingCV());
+                }
+                if (includeTrailingMean || includeTrailingCV)
+                {
+                    dataJsonObject.put("TrailingStartDate", plotData.getTrailingStart());
+                    dataJsonObject.put("TrailingEndDate", plotData.getTrailingEnd());
```
labkey-jeckels: Do we need the end date? Won't that always be the same as the date of the last data point?

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1268,6 +1290,15 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
             me = this;
 
+        let panelY = me.canUserEdit() ? -375 : -270;
+        if (valueName === "TrailingMean" || valueName === "TrailingCV") {
+            panelY = panelY + 200;
+        }
+
+        let trailingRuns = this.trailingRuns;
+        let trailingStartDate = row['TrailingStartDate'];
+        let trailingEndDate = row['TrailingEndDate'];
```
labkey-jeckels: See my earlier comment - I bet this could just use the "regular" date for the row.

webapp/TargetedMS/js/TrailingMeanPlotHelper.js
labkey-jeckels: @wendyinnis reminder

labkey-jeckels (approved):
A couple of followup items but no need for additional review.

src/org/labkey/targetedms/model/QCPlotFragment.java
ankurjuneja: no, we don't. removed

PR #666, opened 2023-01-27
Cast row values to match new API behavior

labkey-tchad:
#### Rationale
Floating point values tend to be returned as `BigDecimal` after the latest remote API update.

#### Related Pull Requests
https://github.com/LabKey/server/pull/390

#### Changes
* Don't assume numerical values API `Row`s will be `Double`s

### DIFF
--- a/test/src/org/labkey/test/tests/targetedms/TargetedMSDocumentFormatsTest.java
+++ b/test/src/org/labkey/test/tests/targetedms/TargetedMSDocumentFormatsTest.java
@@ -167,8 +167,8 @@ public class TargetedMSDocumentFormatsTest extends TargetedMSTest
         int index = 0;
         for (Row row : response.getRowset())
         {
-            Assert.assertEquals("Wrong Mz for precursor " + index, precursorMz, (Double)row.getValue("PrecursorId/Mz"), 0.0001);
-            Assert.assertEquals("Wrong Mz for transition " + index, transitionMz[index++], (Double)row.getValue("Mz"), 0.0001);
+            Assert.assertEquals("Wrong Mz for precursor " + index, precursorMz, ((Number)row.getValue("PrecursorId/Mz")).doubleValue(), 0.0001);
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-adam I imagine we should capture this behavior change in our release notes? Unlike many of the other changes, it won't be caught by the compiler.

https://github.com/LabKey/labkey-api-java/blob/develop/CHANGELOG.md

labkey-adam:
> @labkey-adam I imagine we should capture this behavior change in our release notes? Unlike many of the other changes, it won't be caught by the compiler.
> 
> https://github.com/LabKey/labkey-api-java/blob/develop/CHANGELOG.md

@labkey-jeckels I disagree. This was an unintentional change that should be reverted. I tracked it down... we stopped passing `Command` into `CommandResponse` constructors and stashing requiredVersion, because this was almost never used. We kept it in the one place where it was used, `RowsResponse()`, but `_requiredVersion` was set after `fixupParsedData()` was called. As a result, the fixup didn't know this was an enhanced response. Just need to move one line. I'll open a PR and publish 5.0.1.

@labkey-tchad I suggest we close this PR and fix by switching to v5.0.1.

labkey-jeckels:
> > @labkey-adam I imagine we should capture this behavior change in our release notes? Unlike many of the other changes, it won't be caught by the compiler.
> > https://github.com/LabKey/labkey-api-java/blob/develop/CHANGELOG.md
> 
> @labkey-jeckels I disagree. This was an unintentional change that should be reverted. I tracked it down... we stopped passing `Command` into `CommandResponse` constructors and stashing requiredVersion, because this was almost never used. We kept it in the one place where it was used, `RowsResponse()`, but `_requiredVersion` was set after `fixupParsedData()` was called. As a result, the fixup didn't know this was an enhanced response. Just need to move one line. I'll open a PR and publish 5.0.1.
> 
> @labkey-tchad I suggest we close this PR and fix by switching to v5.0.1.

Sticking with greater backwards compatibility approach sounds good to me.

labkey-tchad:
This was a remoteapi bug

PR #674, opened 2023-02-09
Fix testSamplesExcludedForSingleMetric - correct plotcounts on load

ankurjuneja:


labkey-jeckels:
This feels like we're probably getting different results based on the individual test cases running in a different order, and whatever is running prior to testSamplesExcludedForSingleMetric() is leaving the config in a plot-per-peptide setup instead of a combined plot.

@labkey-sweta do you agree? And if so, can we either make the order deterministic or make sure that each test case configures the plot as desired instead of using whatever was last set?

labkey-jeckels:
@labkey-sweta and I talked this afternoon. She explained that while the test used to wait for at least N plots before proceeding, now it waits for exactly N. So we believe it's always been seeing 7 plots.

labkey-jeckels (approved):

PR #675, opened 2023-02-10
Automation test for Access mean and CV cross last N runs

labkey-sweta:
#### Rationale
Automation test for Access mean and CV cross last N runs

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
@@ -176,8 +176,7 @@ public final class QCPlotsWebPart extends BodyWebPart<QCPlotsWebPart.Elements>
 
     public void setTrailingLast(String value)
     {
-        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().trailingLast));
-        getWrapper().setFormElement(elementCache().trailingLast, value);
+        doAndWaitForUpdate(() -> getWrapper().setFormElement(elementCache().trailingLast, value));
     }
 
     public List<String> getMetricTypeOptions()
--- a/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
@@ -169,6 +169,17 @@ public final class QCPlotsWebPart extends BodyWebPart<QCPlotsWebPart.Elements>
         }
     }
 
+    public String getTrailingLast()
+    {
+        return getWrapper().getFormElement(elementCache().trailingLast);
+    }
+
+    public void setTrailingLast(String value)
+    {
+        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().trailingLast));
+        getWrapper().setFormElement(elementCache().trailingLast, value);
+    }
+
     public List<String> getMetricTypeOptions()
     {
         return getWrapper()._ext4Helper.getComboBoxOptions(elementCache().metricTypeCombo);
@@ -399,6 +410,11 @@ public final class QCPlotsWebPart extends BodyWebPart<QCPlotsWebPart.Elements>
             throw new NoSuchElementException("Unable to find svg point with with acquired date: " + dateStr);
         }
         return point;
+    }
+
+    public int getTotalPlotCount()
+    {
+        return elementCache().findPlots().size();
     }
 
     public WebElement openExclusionBubble(String acquiredDate)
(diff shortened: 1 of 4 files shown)


test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
```diff
@@ -169,6 +169,17 @@ private void toggleQCPlotTypes(Set<QCPlotType> plotTypes)
...
 
+    public String getTrailingLast()
+    {
+        return getWrapper().getFormElement(elementCache().trailingLast);
+    }
+
+    public void setTrailingLast(String value)
+    {
+        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().trailingLast));
+        getWrapper().setFormElement(elementCache().trailingLast, value);
```
labkey-tchad: Wait for plots to update.
```suggestion
        doAndWaitForUpdate(() -> getWrapper().setFormElement(elementCache().trailingLast, value));
```

labkey-tchad (changes requested):
Mean and CV plots should contain fewer points than a normal plot, then even fewer than that when you add a guide set (or sets). Checking those would be a good proxy to verify that the special calculations are getting applied without actually checking the values.

test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
labkey-sweta: done

test/src/org/labkey/test/tests/targetedms/TargetedMSTrailingMeanAndCVTest.java
```diff
@@ -0,0 +1,177 @@
...
+                "TrailingCV - Please enter a positive integer (>2) that is less than or equal to total number of available runs - 47"));
+
+        qcPlotsWebPart.setTrailingLast("48");
+        longWait().until(ExpectedConditions.textToBePresentInElement(Locators.labkeyError.findWhenNeeded(getDriver()),
+                "TrailingCV - The number you entered is larger than the number of available runs. Only 47 runs are used for calculation"));
+
+        qcPlotsWebPart.setTrailingLast("3");
+        log("Verifying tooltips");
+        qcPlotsWebPart.waitForPlots(7);
+        sleep(1000);
```
labkey-tchad: What is this sleep waiting for? Does it need to wait for the point? Sleeps should have an explaination.

test/src/org/labkey/test/tests/targetedms/TargetedMSTrailingMeanAndCVTest.java
```diff
@@ -0,0 +1,177 @@
...
+        qcPlotsWebPart.setMetricType(QCPlotsWebPart.MetricType.MASSACCURACY);
+        Assert.assertTrue("Y axis is not labeled correctly for " + QCPlotsWebPart.MetricType.MASSACCURACY,
+                qcPlotsWebPart.getSVGPlotText("precursorPlot0").contains("PPM"));
+
+        qcPlotsWebPart.setMetricType(QCPlotsWebPart.MetricType.RETENTION);
+        Assert.assertTrue("Y axis is not labeled correctly for " + QCPlotsWebPart.MetricType.RETENTION,
+                qcPlotsWebPart.getSVGPlotText("precursorPlot0").contains("Minutes"));
+
+        log("Verifying the count of points on the plot");
+        Assert.assertEquals("Invalid point count for all replicates", 315 , qcPlotsWebPart.getPointElements("d", SvgShapes.CIRCLE.getPathPrefix(), true).size());
```
labkey-tchad: This should explain how 315 is calculated. This sort of assertion can be tricky to reverse-engineer if there is ever a failure.
e.g. `7 plots with 45 points per plot`
Better yet, all of these values should be constants and variables that can be used to calculate expectations, making only once spot to update if the underlying data changes.
e.g. `int expected = REPLICATE_COUNT * (RUN_COUNT - trailingLast + 1)`

The same goes for the other row count assertions.

test/src/org/labkey/test/tests/targetedms/TargetedMSTrailingMeanAndCVTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+        log("Verifying error checking on trailing last number");
+
+        qcPlotsWebPart.setTrailingLast("-1");
+        longWait().until(ExpectedConditions.textToBePresentInElement(Locators.labkeyError.findWhenNeeded(getDriver()),
+                "TrailingCV - Please enter a positive integer (>2) that is less than or equal to total number of available runs - 47"));
+
+        qcPlotsWebPart.setTrailingLast("48");
+        longWait().until(ExpectedConditions.textToBePresentInElement(Locators.labkeyError.findWhenNeeded(getDriver()),
+                "TrailingCV - The number you entered is larger than the number of available runs. Only 47 runs are used for calculation"));
```
labkey-tchad: We should assert that these errors don't prevent other plots from showing up. I think that was an issue during manual testing.

labkey-tchad (approved):
A couple of suggestions. Feel free to request a re-review when you're done or just merge after updating.

test/src/org/labkey/test/tests/targetedms/TargetedMSTrailingMeanAndCVTest.java
labkey-sweta: what is RUN_COUNT?

labkey-tchad: Those were just hypothetical constants.
e.g.
```
private static final RUN_COUNT = 47;
private static final REPLICATE_COUNT = 7;
```

test/src/org/labkey/test/tests/targetedms/TargetedMSTrailingMeanAndCVTest.java
labkey-sweta: Not needed, thank you for pointing out

PR #681, opened 2023-03-04
Issue 47097: Extra guide sets shown when checking Show Reference Guidset option

ankurjuneja:
#### Rationale
Extra guideset is shown when checking Show reference guideset option because of the bug when comparing dates from the custom date range selected by the user and guideset training range.

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/343

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -1420,10 +1420,10 @@ public class TargetedMSController extends SpringActionController
                 Predicate<RawMetricDataSet> withInDateRange = rawMetricDataSet -> rawMetricDataSet.getSampleFile().getAcquiredTime() != null &&
                         (qcStartDate != null &&
                         (rawMetricDataSet.getSampleFile().getAcquiredTime().after(qcStartDate)
-                                || DateUtil.getDateOnly(rawMetricDataSet.getSampleFile().getAcquiredTime()).compareTo(qcStartDate) == 0)) &&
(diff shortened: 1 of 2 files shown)


src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -1464,16 +1464,30 @@ private GuideSet getClosestPastGuideSet(List<GuideSet> guideSets, Date startDate
...
                 {
-                    if (null == gs)
+                    // show reference guideset when
+                    // 1. startDate is after guideset
+                    // 2. startDate is between guideset date range
+                    // 3. startDate is same as guideset training start
+                    // 4. startDate is same as guideset training end
+                    boolean startDateAfterGuideSet = guideSet.getTrainingEnd().before(startDate);
+                    boolean startDateBetweenGuideSet = guideSet.getTrainingStart().before(startDate) && guideSet.getTrainingEnd().after(startDate);
+                    boolean startDateOverlapGuideSetStart = DateUtil.getDateOnly(guideSet.getTrainingStart()).compareTo(DateUtil.getDateOnly(startDate)) == 0;
```
labkey-jeckels: `compareTo() == 0` is effectively `equals()`, isn't it?

labkey-jeckels (approved):
I did some light testing and this works better. There's a little weirdness when you have multiple runs on a single day and start a new guide set mid-day, as you see all of the samples from that day, and not just those that are included in the guide set. But that's probably fine in practice.

src/org/labkey/targetedms/TargetedMSController.java
ankurjuneja: yes

PR #708, opened 2023-04-24
Issue 47774: Remove TargetedMS module properties related to reading chromatogram data

vagisha:
#### Rationale
We can remove the following modules properties since we want to always read chromatogram data from skyd files when they are available:
SKIP_CHROMATOGRAM_IMPORT_PROPERTY
PREFER_SKYD_FILE_CHROMATOGRAMS_PROPERTY

Make MAX_TRANSITION_CHROM_INFOS_PROPERTY and MAX_PRECURSORS_PROPERTY  configurable only at the site level.  Removing these properties breaks TargetedMSPrecursorLevelDataTest. This test (https://github.com/LabKey/targetedms/pull/396) changes the values of these two properties to 0 to test that calibration curves and fold changes can be calculated  even if there is nothing in the TransitionChromInfo table.


labkey-jeckels (approved):
Looks good. Thank you!

PR #737, opened 2023-06-15
Timeout increase

labkey-sweta:
#### Related Pull Requests
* 

#### Changes
*


test/src/org/labkey/test/tests/panoramapremium/TargetedMSHidePeptidesAndMolecules.java
```diff
@@ -14,7 +14,7 @@
 import java.util.Arrays;
 
 @Category({})
-@BaseWebDriverTest.ClassTimeout(minutes = 3)
+@BaseWebDriverTest.ClassTimeout(minutes = 5)
```
labkey-danield: Why not make it 6 just to give a little extra buffer?

labkey-danield (approved):

PR #742, opened 2023-06-20
Support for skyd file format 18 needed for Skyline 23.1

nickshulman:
#### Rationale
Skyline 23.1 will have some changes to the .skyd file format which will need to be supported by Panorama

#### Changes
The layout of the structure "ChromGroupHeaderInfo" changed a lot in the latest .skyd file format, so there's a big "if" branch in the ChromGroupHeaderInfo constructor that reads the fields in different order.
Instead of just having a string for each chromatogram TextId, there is a protocol buffer called "ChromatogramGroupId" which has separate fields for the peptide modified sequence and all the small molecule fields.
The reason that ChromatogramGroupId was introduced in Skyline is so that precursors with different "spectrum filters" can be distinguished in the .skyd file. This pull request currently just ignores any spectrum filters which might be in the .skyd file.


labkey-jeckels:
@nickshulman thanks for these updates.

We're branching early next week for our 23.7 release, which is scheduled for mid-July. When do you expect the skyd updates to be finalized? I imagine we'll want these changes in 23.7, whether it's in the initial release or a maintenance release.

labkey-jeckels:
@nickshulman - anything we can help with on the LabKey side so that Panorama is ready when the new Skyline is released? I'm not sure on the target date.

src/org/labkey/targetedms/parser/ChromatogramGroupId.java
```diff
@@ -0,0 +1,66 @@
...
+    {
+        ChromatogramGroupId chromatogramGroupId = new ChromatogramGroupId();
+        chromatogramGroupId._qcTraceName = qcTraceName;
+        return chromatogramGroupId;
+    }
+
+    public static List<ChromatogramGroupId> fromProtos(ChromatogramGroupDataOuterClass.ChromatogramGroupIdsProto proto)
+    {
+        List<Target> targets = new ArrayList<>();
+        targets.add(null);
```
labkey-jeckels, line 57: I assume this is to make the `targets` list one-based?

```suggestion
        // Make one-based lookups easy
        targets.add(null);
```

labkey-jeckels (approved):
This looks good to me. Assuming that Skyline 23.1 will be released before mid-November,  we will want to backport this into 23.7 as well. We can do that after merging this PR to develop.

PR #745, opened 2023-06-21
Automation test for holistic instrument-tracking folders

labkey-sweta:
#### Rationale
Automation test for holistic instrument-tracking folders

https://docs.google.com/document/d/1lpbZNSWnOpZJU7ELgHROS_a7AR4CE9AASC0RNWFXfwM/edit#heading=h.16635yt5s9h

#### Related Pull Requests
* 

#### Changes
New component class: UtilizationCalendarWebPart.java
New test class: TargetedMSUtilizationCalendarTest.java

### DIFF
--- a/test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
@@ -77,7 +77,7 @@ public class UtilizationCalendarWebPart extends BodyWebPart<UtilizationCalendarW
         if (endDate != null)
             getWrapper().setFormElement(Locator.name("event-end-date").findElement(modalDialog.getComponentElement()), endDate);
         modalDialog.dismiss("Save", 0);
-        getWrapper().shortWait().until(ExpectedConditions
+        getWrapper().longWait().until(ExpectedConditions
                 .visibilityOfNestedElementsLocatedBy(modalDialog.getComponentElement(), Locators.labkeyError));
         Assert.assertEquals("Incorrect validation message", errorMsg, Locators.labkeyError.findElement(modalDialog.getComponentElement()).getText());
         modalDialog.dismiss("Cancel", 0);
@@ -112,7 +112,6 @@ public class UtilizationCalendarWebPart extends BodyWebPart<UtilizationCalendarW
     {
         String day = date.substring(date.lastIndexOf("-") + 1, date.length());
         getWrapper().log("Clicking " + day + " to update the status");
-        getWrapper().mouseOut();
         getWrapper().sleep(500);
         getWrapper().mouseOver(elementCache().dayLoc.withText(day).findElement(elementCache().calendar.findWhenNeeded(this)));
         return getWrapper().waitForElement(elementCache().tooltip).getText();
--- a/test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
@@ -1,12 +1,14 @@
 package org.labkey.test.components.targetedms;
 
 import org.jetbrains.annotations.Nullable;
+import org.junit.Assert;
 import org.labkey.test.Locator;
 import org.labkey.test.Locators;
 import org.labkey.test.components.BodyWebPart;
 import org.labkey.test.components.bootstrap.ModalDialog;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
 
 public class UtilizationCalendarWebPart extends BodyWebPart<UtilizationCalendarWebPart.Elements>
 {
@@ -34,7 +36,6 @@ public class UtilizationCalendarWebPart extends BodyWebPart<UtilizationCalendarW
     {
         getWrapper().doAndWaitForElementToRefresh(() -> getWrapper().selectOptionByValue(elementCache().display, value),
                 elementCache().yearTitle, getWrapper().shortWait());
-
         return this;
     }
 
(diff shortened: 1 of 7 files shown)


test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
```diff
@@ -0,0 +1,122 @@
...
+    {
+        ModalDialog offlineAnnotationDialog = clickDate(startDate);
+        getWrapper().setFormElement(Locator.name("event-description").findElement(offlineAnnotationDialog.getComponentElement()), description);
+        getWrapper().setFormElement(Locator.name("event-start-date").findElement(offlineAnnotationDialog.getComponentElement()), startDate);
+        getWrapper().setFormElement(Locator.name("event-end-date").findElement(offlineAnnotationDialog.getComponentElement()), endDate);
+        offlineAnnotationDialog.dismiss("Save", 0);
+        return this;
+    }
+
+    public String getValidationErrorMsg(String startDate, @Nullable String endDate, @Nullable String description)
```
labkey-jeckels: Can this be consolidated with the `markOffline` method? Could it call the other method and then look for the error message?

test/src/org/labkey/test/tests/targetedms/TargetedMSUtilizationCalendarTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private void doInit()
+    {
+        setupFolder(TargetedMSTest.FolderType.QC);
+        importData(SProCoP_FILE);
+    }
+
+    @Test
+    public void testUtilizationCalendarActions()
```
labkey-jeckels, line 40: Please also check that the heat map is showing data from the imported Skyline file, and that toggling the heat map data source between samples and outliers works.

labkey-jeckels (changes requested):

test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
labkey-tchad: Yeah, this should be named something like `markOfflineExpectingError` and consolidated with `markOffline`.
I wouldn't expect a `get*` method to change the page state very much. A getter might open a combo box to inspect its contents or trigger a tooltip but not much more than that.
Also, having tests validate the error message after closing the dialog makes it impossible to collect informative failure screenshots.

test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
```diff
@@ -0,0 +1,122 @@
...
+    public String getDisplay()
+    {
+        return getWrapper().getSelectedOptionText(elementCache().display);
+    }
+
+    public UtilizationCalendarWebPart setHeatMap(String value)
+    {
+        getWrapper().selectOptionByText(elementCache().heatMap, value);
+        return this;
+    }
```
labkey-tchad, line 51: Does this need to wait for the heatmap to update?

test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
```diff
@@ -0,0 +1,122 @@
...
+    public String getValidationErrorMsg(String startDate, @Nullable String endDate, @Nullable String description)
+    {
+        ModalDialog modalDialog = clickDate(startDate);
+        getWrapper().setFormElement(Locator.name("event-start-date").findElement(modalDialog.getComponentElement()), startDate);
+        if (description != null)
+            getWrapper().setFormElement(Locator.name("event-description").findElement(modalDialog.getComponentElement()), description);
+        if (endDate != null)
+            getWrapper().setFormElement(Locator.name("event-end-date").findElement(modalDialog.getComponentElement()), endDate);
+        modalDialog.dismiss("Save", 0);
+        String errorMsg = Locators.labkeyError.findElement(modalDialog.getComponentElement()).getText();
```
labkey-tchad: The test is failing with empty error messages. This should wait for the error to have some text; waiting for visibility should do the trick.
```suggestion
        String errorMsg = getWrapper().shortWait().until(ExpectedConditions
                .visibilityOfNestedElementsLocatedBy(modalDialog.getComponentElement(), Locators.labkeyError).getText();
```

test/src/org/labkey/test/tests/targetedms/TargetedMSUtilizationCalendarTest.java
```diff
@@ -0,0 +1,103 @@
...
+                .gotoUtilizationCalendar();
+        try
+        {
+            utilizationCalendar.clickDate("2013-08-01");
+            Assert.fail("Readers should not be able to update the status");
+        }
+        catch (Exception e)
+        {
+            Assert.assertTrue("Reader cannot update the status", true);
+        }
```
labkey-tchad: This assertion will never fail and looks like a mistake. Just leave a comment explaining that we're expecting an exception. Also, this should only catch the expected exception (maybe `NoSuchElementException`?).
```suggestion
        catch (NoSuchElementException expected)
        {
            // Reader cannot update the status
            // Additional validation?
        }
```

labkey-tchad (changes requested):

test/src/org/labkey/test/tests/targetedms/TargetedMSUtilizationCalendarTest.java
labkey-sweta: Done

test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
labkey-sweta: Did not need it locally, fingers cross for teamcity.

test/src/org/labkey/test/tests/targetedms/TargetedMSUtilizationCalendarTest.java
labkey-sweta: Done

test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
labkey-sweta: Done

labkey-jeckels (approved):
Looks like my requests are addressed. Trey should do a followup review too.

test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
labkey-tchad: I would really prefer to add synchronization proactively, rather than waiting for it to fail when the feature/test has faded from our memory.

labkey-tchad (commented):
Re-request once the tests are passing

labkey-tchad (approved):

PR #753, opened 2023-07-19
Fix missing chromatograms in .skyd files whose format is older than 12

nickshulman:
#### Rationale
Panorama is unable to find chromatograms for documents which were created before 2017.
This is because in older .skyd files, "CacheHeaderStruct.chromTransitionSize" is always zero.
The correct way to get the size would be _cacheFormat.chromTransitionSerializer().getItemSizeOnDisk().

#### Changes

There's probably a better way to implement "countMissingChromatograms" involving "SELECT COUNT" instead of actually counting the number of rows returned, but I could not figure out how to write that code.


test/src/org/labkey/test/tests/targetedms/TargetedMSDocumentFormatsTest.java
```diff
@@ -171,4 +181,17 @@ private void validateCrossLinked(String file, int countWithAnyIon, String modifi
             Assert.assertEquals("Wrong Mz for transition " + index, transitionMz[index++], (Double)row.getValue("Mz"), 0.0001);
         }
     }
+
+    private int countMissingChromatograms(String file) throws Exception
+    {
+        Connection cn = createDefaultConnection();
+
+        List<Filter> filters = List.of(
```
labkey-jeckels, line 189: Here's an example of firing a SQL query in a test:

https://github.com/labkey/targetedms/blob/develop/test/src/org/labkey/test/tests/targetedms/AbstractQuantificationTest.java#L444

It would be more efficient, but we're not loading a ton of data in this test so I'm not sure it would be noticeable.

labkey-jeckels (approved):
We've already branched for 23.7 so this won't hit most Panorama deployments until 23.11 in November. I don't know the scenario that discovered this problem, but we might want to target release23.7-SNAPSHOT instead.

vagisha:
> We've already branched for 23.7 so this won't hit most Panorama deployments until 23.11 in November. I don't know the scenario that discovered this problem, but we might want to target release23.7-SNAPSHOT instead.

I found this issue when I was trying to import old Panorama tutorial files. Some of the documents used in automated TargetedMS tests also have the same problem.  I doubt that users will run into this issue on PanoramaWeb.  But if it is easy to target release23.7-SNAPSHOT instead that would be great.

PR #755, opened 2023-08-07
Backport .skyd file changes from develop

nickshulman:
#### Rationale
Skyline 23.1 will have a new .skyd format

#### Related Pull Requests
#753 #742

#### Changes
The layout of the structure "ChromGroupHeaderInfo" changed a lot in the latest .skyd file format, so there's a big "if" branch in the ChromGroupHeaderInfo constructor that reads the fields in different order.
Instead of just having a string for each chromatogram TextId, there is a protocol buffer called "ChromatogramGroupId" which has separate fields for the peptide modified sequence and all the small molecule fields.
The reason that ChromatogramGroupId was introduced in Skyline is so that precursors with different "spectrum filters" can be distinguished in the .skyd file. This pull request currently just ignores any spectrum filters which might be in the .skyd file.


nickshulman:
Oops. Accidentally named the branch "23.3" instead of "23.7"

PR #766, opened 2023-09-15
Issue 48689: Panorama fails to parse .skyd version 17 files

labkey-jeckels:
#### Rationale
Updates to support new .skyd file formats didn't map the right parsing code to version 17.

#### Changes
* Treat version 17 like version 16 (and others) for the header parsing


nickshulman:
Looks good to me. I would like to try to add a test case to TargetedMSDocumentFormatsTest for this.

The only Skyline-daily which used .skyd format 17 was 22.2.1.501 which can be installed from here:
https://skyline.ms/_webdav/home/software/Skyline/daily/%40files/Skyline-daily-64_22_2_1_501.zip

I will try to use that version of Skyline-daily to modify one of our existing test documents so that the skyd format is 17.

labkey-jeckels:
@nickshulman that would be great. My seldom-used Windows desktop is refusing to boot at the moment so I can't easily hack up a test file. If you can create one, I can add coverage for it.

nickshulman (approved):

labkey-jeckels:
Thanks Nick!

PR #777, opened 2023-10-12
Ticket 48872: Unable to customize columns in PTM reports

labkey-jeckels:
#### Rationale
Users are generally satisfied by the default setup for MAM reports. But sometimes they want to customize the display a bit more

#### Changes
* Scope each run's reporting to a query that allows for its own customization of columns
* Expose `AminoAcid` and `Location` as separate columns, as well as the concatenated `SiteLocation`
* Revert to single-row display in the early stage PTM report when the user customizes the sort in a way that's not reliably grouped together

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSSchema.java
+++ b/src/org/labkey/targetedms/TargetedMSSchema.java
@@ -1742,35 +1742,33 @@ public class TargetedMSSchema extends UserSchema
 
         if (result == null && StringUtils.startsWithIgnoreCase(queryName, QUERY_PTM_PERCENTS_PREFIX))
         {
-            String runIdString = queryName.substring(QUERY_PTM_PERCENTS_PREFIX.length());
-            try
-            {
-                int runId = Integer.parseInt(runIdString);
-                QueryDefinition queryDef = Objects.requireNonNull(getQueryDef("PTMPercents"));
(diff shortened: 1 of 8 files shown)


src/org/labkey/targetedms/TargetedMSSchema.java
```diff
@@ -1734,6 +1735,43 @@ public Set<String> getTableNames()
...
+    public QueryDefinition getQueryDef(@NotNull String queryName)
+    {
+        QueryDefinition result = super.getQueryDef(queryName);
+
+        if (result == null && StringUtils.startsWithIgnoreCase(queryName, QUERY_PTM_PERCENTS_PREFIX))
+        {
+            String runIdString = queryName.substring(QUERY_PTM_PERCENTS_PREFIX.length());
+            try
+            {
+                int runId = Integer.parseInt(runIdString);
```
ankurjuneja: runIds are long type in targetedms

ankurjuneja (approved):

src/org/labkey/targetedms/TargetedMSSchema.java
labkey-jeckels: True. It'll be a while before we hit 2 billion Skyline files imported, but best to use the data type consistently. Pushed an update.

PR #778, opened 2023-10-12
Dismiss old tooltip before triggering a new one

labkey-tchad:
#### Rationale
Test sometimes has two tooltips showing. Need to try not to trigger multiple tooltips.

#### Related Pull Requests
* N/A

#### Changes
* Dismiss old tooltip before triggering a new one
* Clean up some IntelliJ warnings

### DIFF
--- a/test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
@@ -3,7 +3,7 @@ package org.labkey.test.components.targetedms;
 import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
 import org.labkey.test.Locator;
-import org.labkey.test.Locators;
+import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.BodyWebPart;
 import org.labkey.test.components.bootstrap.ModalDialog;
 import org.openqa.selenium.WebDriver;
@@ -22,7 +22,7 @@ public class UtilizationCalendarWebPart extends BodyWebPart<UtilizationCalendarW
 
     public void waitForPage()
     {
-        getWrapper().waitFor(() -> elementCache().yearTitle.isDisplayed(this),
+        WebDriverWrapper.waitFor(() -> elementCache().yearTitle.isDisplayed(this),
                 "Calendar did not load on time",
                 getWrapper().defaultWaitForPage);
     }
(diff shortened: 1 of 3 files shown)


labkey-sweta (commented):
Test is still failing

test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
```diff
@@ -22,7 +22,7 @@ public UtilizationCalendarWebPart(WebDriver driver)
 
     public void waitForPage()
     {
-        getWrapper().waitFor(() -> elementCache().yearTitle.isDisplayed(this),
+        WebDriverWrapper.waitFor(() -> elementCache().yearTitle.isDisplayed(this),
                 "Calendar did not load on time",
```
labkey-sweta, line 26: What is the difference between getWrapper wait and WebDriverWrapper wait?

test/src/org/labkey/test/tests/targetedms/TargetedMSUtilizationCalendarTest.java
```diff
@@ -20,7 +20,7 @@ public class TargetedMSUtilizationCalendarTest extends TargetedMSTest
     @Override
     protected @Nullable String getProjectName()
     {
-        return getCurrentTestClass().getSimpleName() + " Project";
+        return TargetedMSUtilizationCalendarTest.class.getSimpleName() + " Project";
     }
```
labkey-sweta, line 24: Curious, how is it different?

test/src/org/labkey/test/components/targetedms/UtilizationCalendarWebPart.java
labkey-tchad: `WebDriverWrapper.waitFor` is a static method. Referencing it through the non-static `getWrapper` is allowed in Java but is misleading.

test/src/org/labkey/test/tests/targetedms/TargetedMSUtilizationCalendarTest.java
labkey-tchad: In reality, it probably doesn't ever evaluate to something different but `getCurrentTestClass` is a static method in BaseWebDriverTest that could return `null`.
`getClass()` is actually the method that this should have used (I just changed it).

labkey-tchad:
Unrelated tests are failing now

labkey-sweta (approved):

PR #781, opened 2023-11-01
Sample getMaterialDetailsBaseURL to use the sample container instead of the sample type container

cnathe:
#### Rationale
The TargetedMSSampleManagerIntegrationTest.testSampleTypeNavigation test case started failing yesterday because the call at line 185 in the SampleFileTable.java class gets the ExpSampleType previously was returning the current container for the `sampleType.getContainer()` call. It is now returning the container where the sample type is defined in, which seems more correct (not sure exactly which PR changed this). The test however is expecting the current container (i.e. the sample's container). This PR fixes that by using the `matchingContainers` var in that file.

#### Changes
* SampleFileTable to use matchingContainer in getMaterialDetailsBaseURL() call


cnathe:
Note that this will be changed to target release23.11-SNAPSHOT once that is available. So you can hold off on CR until then.

PR #782, opened 2023-11-01
Increase length of the CustomIonName columns

vagisha:
#### Rationale
Custom ion names can be really long.  Increase the length of the CustomIonName column in the following tables:
- Molecule
- MoleculePrecursor
- MoleculeTransition
- ExcludedPrecursors


labkey-jeckels (approved):
This change looks good. It will just miss our cutoff for the 23.11 branch, which is 6PM yesterday. I think that's OK as PanoramaWeb is the only server where users have hit the current limit and you've already expanded the columns there. But if you'd like to include this in 23.11, the branches should be created later today and you could recreate the PR.

vagisha:
Replaced with https://github.com/LabKey/targetedms/pull/784

PR #845, opened 2024-03-01
Issue 49427: Inconsistent number of metrics in Panorama QC folder

labkey-jeckels:
#### Rationale
Different QC folders may have different numbers of active QC metrics. We're currently looking at the parent folder to get the count.

#### Changes
* Use the child folder when resolving the active metrics


ankurjuneja (approved):
looks like the target branch is incorrect.

PR #846, opened 2024-03-01
Issue 49687: PrecursorHtmlMaker.getHtml() should return HtmlString

labkey-jeckels:
#### Rationale
HtmlString is more robust

#### Changes
* Use the DOM builder


ankurjuneja (approved):
target branch needs to updated.

labkey-jeckels:
> target branch needs to updated.

I'll update it to 24.3 when the branch is created tomorrow, before I merge. This should run tests overnight against develop.

PR #857, opened 2024-03-13
Automation test for Issue 49427: Inconsistent number of metrics in Panorama QC folder 

labkey-sweta:
#### Rationale
Automation test for Issue 49427: Inconsistent number of metrics in Panorama QC folder 

Repro steps 
1. Create a Panorama QC folder.
2. Create a subfolder that's also a QC folder.
3. Load a Skyline document into the subfolder.
4. Create another subfolder (sibling to the other subfolder) that's also a QC folder.
5. Load the same Skyline document into the second subfolder.
6. In the second subfolder, disable one or more of the QC metrics.
7. Go to the parent folder.
8. Check that the subfolders are reporting different number of metrics.

#### Related Pull Requests
* 

#### Changes
Added new test class TargetedMSQCConfigureMetricTest.
Added new ConfigureQCMetricsWebPart component.


test/src/org/labkey/test/tests/targetedms/TargetedMSQCConfigureMetricTest.java
```diff
@@ -0,0 +1,94 @@
...
+        new PanoramaDashboard(this).getQcSummaryWebPart().clickMenuItem("Configure QC Metrics");
+        configureQCMetrics = new ConfigureQCMetricsWebPart(getDriver());
+        configureQCMetrics.configureMetric("Precursor Area", "Disabled")
+                .clickSave();
+
+        log("Verify the data is displayed correctly on parent folder");
+        goToProjectHome();
+        QCSummaryWebPart qcSummary = new PanoramaDashboard(this).getQcSummaryWebPart();
+        List<QCSummaryWebPart.QcSummaryTile> subFolderTile = qcSummary.getQcSummaryTiles();
+        if (subFolderTile.get(1).getFolderName().equals(SUBFOLDER_1))
```
labkey-jeckels: We should be ordering these folders consistently so I don't think we should need the `if` check here. Were they rendering in different orders for you? If so, I'd like to fix that so it's not random for users.

test/src/org/labkey/test/tests/targetedms/TargetedMSQCConfigureMetricTest.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    protected String getProjectName()
+    {
+        return "TargetedMSQCConfigureMetricTest Project";
+    }
+
+    @Override
+    public List<String> getAssociatedModules()
+    {
+        return Arrays.asList();
```
labkey-jeckels, line 85: Is this doing anything useful? I would expect we'd want to say it's `targetedms` like the superclass does.

test/src/org/labkey/test/components/targetedms/QCSummaryWebPart.java
```diff
@@ -176,6 +177,10 @@ private void readSummary()
             Pattern precursorPattern = Pattern.compile("(\\d+) precursor");
             Matcher precursorMatcher = precursorPattern.matcher(qcSummary);
             _precursorCount = precursorMatcher.find() ? Integer.parseInt(precursorMatcher.group(1)) : 0;
+
+            Pattern metricPattern = Pattern.compile("(\\d+) metric");
```
labkey-jeckels, line 181: This should be fine, but I'm always happy to tag the DOM with IDs or classes to help tests find the right parts.

labkey-jeckels (commented):
Looks good in terms of coverage. A couple of questions, and I'll let Trey review for consistency with overall conventions.

test/src/org/labkey/test/tests/targetedms/TargetedMSQCConfigureMetricTest.java
labkey-sweta: The reason for adding if condition was because we QCSummaryTiles is returning list and order cannot be determined.
The display of tiles on the UI was consistent. 
@labkey-jeckels let me know if you want me to it to Set instead of List.

test/src/org/labkey/test/components/targetedms/QCSummaryWebPart.java
labkey-sweta: This is seems to be okay for now. Thank you

test/src/org/labkey/test/tests/targetedms/TargetedMSQCConfigureMetricTest.java
labkey-sweta: My understanding is getAssociatedModules does not do much, but will ask the team and update this accordingly.

test/src/org/labkey/test/tests/targetedms/TargetedMSQCConfigureMetricTest.java
labkey-jeckels: I would expect that the List would consistently reflect the order of the tiles in the UI, so it would be safe to validate/assume which index they're at.

test/src/org/labkey/test/components/targetedms/ConfigureQCMetricsWebPart.java
```diff
@@ -0,0 +1,62 @@
+package org.labkey.test.components.targetedms;
+
+import org.labkey.test.Locator;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class ConfigureQCMetricsWebPart extends WebDriverComponent<ConfigureQCMetricsWebPart.ElementCache>
```
labkey-tchad: I think we already have a class for this: `ConfigureMetricsUIPage`

test/src/org/labkey/test/tests/targetedms/TargetedMSQCConfigureMetricTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.labkey.test.components.targetedms.ConfigureQCMetricsWebPart;
+import org.labkey.test.components.targetedms.QCSummaryWebPart;
+import org.labkey.test.pages.targetedms.PanoramaDashboard;
+
+import java.util.Arrays;
+import java.util.List;
+
+@Category({})
+@BaseWebDriverTest.ClassTimeout(minutes = 2)
+public class TargetedMSQCConfigureMetricTest extends TargetedMSTest
```
labkey-tchad: If this test extends `TargetedMSPremiumTest` instead, it would have access to some QC metric related methods.

labkey-tchad (changes requested):
I think some of these helper changes are duplicating existing functionality in `TargetedMSPremiumTest` and `ConfigureMetricsUIPage`

labkey-tchad (approved):

PR #859, opened 2024-03-13
New fixed deviation from mean outlier option

ankurjuneja:
#### Rationale
This PR adds a new system suitability metric to identify outliers when they deviate from the mean by a fixed amount.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5323

### DIFF
--- a/src/org/labkey/targetedms/model/QCPlotFragment.java
+++ b/src/org/labkey/targetedms/model/QCPlotFragment.java
@@ -117,10 +117,6 @@ public class QCPlotFragment
             {
                 dataJsonObject.put("ValueOutlier", true);
             }
-            if (plotData.isMeanDeviationOutlier(targetedStats.get(plotData.getGuideSetKey())))
-            {
-                dataJsonObject.put("MeanDeviationOutlier", true);
-            }
             dataJsonObject.put("SampleFileId", plotData.getSampleFile().getId());
             dataJsonObject.put("PrecursorChromInfoId", plotData.getPrecursorChromInfoId());
             boolean ignoreInQC = plotData.getSampleFile().isIgnoreInQC(plotData.getMetricId());
--- a/src/org/labkey/targetedms/model/QCPlotFragment.java
+++ b/src/org/labkey/targetedms/model/QCPlotFragment.java
@@ -117,6 +117,10 @@ public class QCPlotFragment
             {
                 dataJsonObject.put("ValueOutlier", true);
             }
+            if (plotData.isMeanDeviationOutlier(targetedStats.get(plotData.getGuideSetKey())))
+            {
+                dataJsonObject.put("MeanDeviationOutlier", true);
+            }
             dataJsonObject.put("SampleFileId", plotData.getSampleFile().getId());
             dataJsonObject.put("PrecursorChromInfoId", plotData.getPrecursorChromInfoId());
             boolean ignoreInQC = plotData.getSampleFile().isIgnoreInQC(plotData.getMetricId());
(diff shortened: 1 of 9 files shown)


src/org/labkey/targetedms/model/QCPlotFragment.java
```diff
@@ -117,6 +117,10 @@ public JSONObject toJSON(boolean includeLJ, boolean includeMR, boolean includeMe
             {
                 dataJsonObject.put("ValueOutlier", true);
             }
+            if (plotData.isMeanDeviationOutlier(targetedStats.get(plotData.getGuideSetKey())))
```
labkey-jeckels: Do we need to return this separate from other value outliers? We're currently using `ValueOutlier` for fixed cutoff or Levey-Jennings, so it seems unexpected to break out mean deviation.

src/org/labkey/targetedms/model/RawMetricDataSet.java
```diff
@@ -356,6 +356,34 @@ public boolean isValueOutlier(GuideSetStats stat)
         }
     }
 
+    public boolean isMeanDeviationOutlier(GuideSetStats stat)
```
labkey-jeckels: See previous comment. I think this can be folded in as a third case in `isValueOutlier()`

src/org/labkey/targetedms/query/QCEnabledMetricsTable.java
```diff
@@ -101,6 +101,21 @@ private static void validateBounds(Map<String, Object> row) throws ValidationExc
...
+                {
+                    if (upperBound == null)
+                    {
+                        throw new ValidationException("For Mean Deviation Cut-Off configuration, you must provide an upper bound");
+                    }
+                    if (lowerBound == null)
+                    {
+                        throw new ValidationException("For Mean Deviation Cut-Off configuration, you must provide a lower bound");
+                    }
+                    if (lowerBound >= 0)
```
labkey-jeckels, line 118: I guess we should check that the `upperBound` is > 0 too.

webapp/TargetedMS/js/LeveyJenningsPlotHelper.js
```diff
@@ -130,8 +130,15 @@ Ext4.define("LABKEY.targetedms.LeveyJenningsPlotHelper", {
...
-        plotProperties['boundType'] = metricProps.metricStatus === LABKEY.targetedms.MetricStatus.ValueCutoff ? 'absolute' : 'stddev'
-
+        if (metricProps.metricStatus === LABKEY.targetedms.MetricStatus.ValueCutoff) {
+            plotProperties['boundType'] = 'absolute';
+        }
+        else if (metricProps.metricStatus === LABKEY.targetedms.MetricStatus.MeanDeviationCutoff) {
+            plotProperties['boundType'] = 'meanDev';
+        }
+        else {
+            plotProperties['boundType'] = 'stddev';
```
labkey-jeckels: See comment in related PR about making these constants.

webapp/TargetedMS/js/LeveyJenningsPlotHelper.js
```diff
@@ -162,7 +169,13 @@ Ext4.define("LABKEY.targetedms.LeveyJenningsPlotHelper", {
             data['value'] = row['Value'];
         }
 
-        data.LJShape = (row.IgnoreInQC ? 'Exclude' : 'Include') + (row.ValueOutlier ? '-Outlier' : '');
+        data.LJShape = (row.IgnoreInQC ? 'Exclude' : 'Include');
+        if (metricProps.metricStatus === LABKEY.targetedms.MetricStatus.MeanDeviationCutoff) {
```
labkey-jeckels: Per earlier comments, can this be collapsed with `ValueOutlier`?

labkey-jeckels (changes requested):
Overall looks good and very small (which is great). A few suggestions/requests.

labkey-jeckels (approved):

PR #868, opened 2024-03-19
New fixed deviation from mean outlier option (#859)

ankurjuneja:
#### Rationale
See related

#### Related Pull Requests
* #859 

#### Changes
*


labkey-jeckels (approved):
I think the test suite kicked off before you back ported the platform change. I've dequeued another test run

PR #870, opened 2024-03-20
Remove parsing of <results_data> element

vagisha:
#### Rationale
<results_data>is not in the Skyline document schema.  According to Nick, it may have been part of an interim build but never made it to the trunk.


labkey-jeckels (approved):
Thanks - I spent some time recently trying unsuccessfully to find a Skyline doc that had this element.

nickshulman (approved):
Looks good to me.

PR #895, opened 2024-05-05
Issue 50011: Outliers for Fixed Deviation from mean for some metric is incorrect

ankurjuneja:
#### Rationale
Outliers for fixed deviation from mean/value is incorrect for metrics that are part of a paired config. We were incorrectly mapping metric configuration of stand alone metrics that are part of a paired config i.e. for the stand alone metric we were mapping it to the paired metric configurations instead of their own configuration that was causing incorrect calculations of stand alone metric outliers.

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/431

#### Changes
*

### DIFF
--- a/src/org/labkey/targetedms/outliers/OutlierGenerator.java
+++ b/src/org/labkey/targetedms/outliers/OutlierGenerator.java
@@ -150,18 +150,18 @@ public class OutlierGenerator
         return sql.toString();
     }
 
-    private Map<String, QCMetricConfiguration> getPreferredMetrics(List<QCMetricConfiguration> configurations, boolean forOutliers)
+    private Map<String, QCMetricConfiguration> getPreferredMetrics(List<QCMetricConfiguration> configurations, boolean forOutlierSummary)
     {
         Map<String, QCMetricConfiguration> preferredConfigs = new LinkedHashMap<>();
         // Deduplicate for metrics that are shown both standalone and paired with another
         for (QCMetricConfiguration configuration : configurations)
         {
             String label1 = configuration.getSeries1Label();
-            retainIfPreferred(preferredConfigs, configuration, label1, configurations, forOutliers);
+            retainIfPreferred(preferredConfigs, configuration, label1, configurations, forOutlierSummary);
             String label2 = configuration.getSeries2Label();
             if (label2 != null)
             {
-                retainIfPreferred(preferredConfigs, configuration, label2, configurations, forOutliers);
+                retainIfPreferred(preferredConfigs, configuration, label2, configurations, forOutlierSummary);
             }
         }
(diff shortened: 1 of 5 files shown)


src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -232,16 +238,35 @@ private String queryContainerSampleFileRawData(List<QCMetricConfiguration> confi
...
-            preferredConfigs.put(label, configuration);
+            if (forOutliers)
+            {
+                if (configuration.getName().equalsIgnoreCase(label))
+                {
+                    preferredConfigs.put(label, configuration);
+                }
+                else
+                {
+                    configurations.forEach(c -> {
```
ankurjuneja: This loop makes it $O(N^2)$. I am open to suggestions to avoid this.

labkey-jeckels: We could switch to passing in a Map (label->config object) with all of the configurations instead of a List.

ankurjuneja: I can check on this but I think that will destroy the original intent of creating this map to avoid the double counting of the outliers for those paired metric configs.

ankurjuneja: This traversal will only happen for the paired metric configs which are just a couple at the moment.

labkey-jeckels: Sorry, I meant switch `configurations` from a List to a Map, not `preferredConfigs` from a Map to a List.

But regardless, given the small number of metrics and the even smaller number that will be subject to this code, it's unlikely to be noticeable in practice.

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -150,28 +150,34 @@ private String getEachSeriesTypePlotDataSql(int seriesIndex, QCMetricConfigurati
         return sql.toString();
     }
 
-    /** @return LabKey SQL to fetch all the values for the specified metrics */
-    private String queryContainerSampleFileRawData(List<QCMetricConfiguration> configurations, Date startDate,
-                                                   Date endDate, List<AnnotationGroup> annotationGroups,
-                                                   boolean showExcluded)
+    private Map<String, QCMetricConfiguration> getPreferredMetrics(List<QCMetricConfiguration> configurations, boolean forOutliers)
```
labkey-jeckels: I suggest renaming `forOutliers` to `forOutlierSummary` since both are used for outliers, just in different ways.

labkey-jeckels (approved):

PR #899, opened 2024-05-08
Ticket 49838 - Early Stage PTM Report updates for Pre-IPA data

labkey-jeckels:
#### Rationale
Users want to leverage the "early-stage" PTM reporting in other tools. That means using a non-pivot format, and not relying on conditional formatting to convey risk.

#### Changes
* Configure the pre-pivoted version of the report with CDR and risk info
* Push the flipping of the percentage for `Lys-loss (Protein C-term K)` into the pre pivot query
* Don't highlight `Gln->pyro-Glu (N-term Q)` at all
* Find the stressed/non-stressed annotation from the right Skyline file

### DIFF
--- a/src/org/labkey/targetedms/query/CDRConditionalFormattingDisplayColumnFactory.java
+++ b/src/org/labkey/targetedms/query/CDRConditionalFormattingDisplayColumnFactory.java
@@ -1,0 +1,196 @@
+package org.labkey.targetedms.query;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.data.ColumnInfo;
+import org.labkey.api.data.CompareType;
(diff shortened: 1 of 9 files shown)


src/org/labkey/targetedms/query/CDRConditionalFormattingDisplayColumnFactory.java
```diff
@@ -0,0 +1,196 @@
...
+
+    public static RiskLevel getRiskLevel(Number value, boolean inCDR, boolean stressed)
+    {
+        if (value == null)
+        {
+            return null;
+        }
+
+        double highCutoff = inCDR ? 0.1 : 0.3;
+        double mediumCutoff = inCDR ? 0.05 : 0.15;
```
ankurjuneja, line 45: would it be good to add a comment about numbers?

ankurjuneja (approved):

PR #920, opened 2024-06-20
Reference webpart name via constant

labkey-adam:
#### Rationale
String references are less than ideal


test/src/org/labkey/test/tests/targetedms/TargetedMSExperimentTest.java
```diff
@@ -204,7 +204,6 @@ protected void verifyProteinSearch()
         //waitForText("1 - 7 of 7");
         assertTextPresentInThisOrder("Protein Search", "Matching Proteins (1)", "Targeted MS Proteins");
 
-        clickAndWait(Locator.id("expandCollapse-ProteinSearchProteinMatches"), 0); // Search results are hidden by default.
```
labkey-jeckels: We should be able to restore this.

labkey-jeckels (approved):

PR #921, opened 2024-06-23
Fix error importing Skyline documents with Spectrum Filters

nickshulman:
#### Rationale
Skyline 23.1 added the "Spectrum Filter" filter feature and some of these documents fail to be imported into Panorama.

#### Changes
"Spectrum Filter" gets parsed from the Skyline document
light:heavy ratios are not calculated for precursors that have a spectrum filter
The spectrum filter is used when deciding which chromatogram to read from the .skyd file for a particular precursor.


src/org/labkey/targetedms/parser/PeakAreaRatioCalculator.java
```diff
@@ -58,6 +58,10 @@ public void init(Map<SkylineDocImporter.SampleFileKey, SampleFile> skylineIdSamp
 
         for(Precursor precursor: _peptide.getPrecursorList())
         {
+            if  (precursor.getSpectrumFilter() != null)
+            {
+                continue;
```
labkey-jeckels, line 66: Can you add a comment describing why we skip if there are spectrum filters?

labkey-jeckels (approved):
Thanks for the code and test cases.

PR #930, opened 2024-07-16
Issue 50773: Pressure traces without a TextId can't be used for QC trace metrics

ankurjuneja:
#### Rationale
Label the trace metrics as  "QC Trace 1", "QC Trace 2", "QC Trace 3".. when sample-scoped pressure traces don't have a TextId value in the SKYD file.


labkey-jeckels (changes requested):
This change looks OK but please target 24.7.

PR #932, opened 2024-07-17
Automation test for Display outliers in heatmaps

labkey-sweta:
#### Rationale
Wrapper for the new web part “Outlier Heatmap  Summary View” which is added to the Panorama QC/SystemSuitability folder. The Web part will display the total number of outliers for each peptide for a user defined period of time. 

New automation test class to test all the scenarios.

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/1991

#### Changes
New test class TargetedMSOutlierHeatmapTest 
New component OutlierHeatmapSummaryWebPart


test/src/org/labkey/test/tests/targetedms/TargetedMSOutlierHeatmapTest.java
```diff
@@ -0,0 +1,139 @@
...
+        _portalHelper.deleteTab("Pareto Plot");
+        _portalHelper.deleteTab("Annotations");
+        _portalHelper.addTab(OUTLIER_HEATMAP_TAB);
+
+        _portalHelper.addWebPart("Outlier Heatmap Summary View");
+        _portalHelper.exitAdminMode();
+    }
+
+    @Test
+    public void testHeatMapColorAndValues() throws InterruptedException
```
labkey-tchad: I don't think anything throws this exception here.
```suggestion
    public void testHeatMapColorAndValues()
```

test/src/org/labkey/test/tests/targetedms/TargetedMSOutlierHeatmapTest.java
```diff
@@ -0,0 +1,139 @@
...
+        log("Verify data range: " + OutlierHeatmapSummaryWebPart.HeatmapDateRange.Last_7_Days);
+        outlierHeatmap.setDateRange(OutlierHeatmapSummaryWebPart.HeatmapDateRange.Last_7_Days, true);
+        Assert.assertEquals("Incorrect outlier count for " + QCPlotsWebPart.MetricType.FWHM, "1",
+                outlierHeatmap.getCellElement(1, QCPlotsWebPart.MetricType.FWHM).getText());
+
+        log("Verify heatmap colors");
+        Assert.assertTrue("Incorrect heatmap color for highest(Red)",
+                outlierHeatmap.getCellElement(1, QCPlotsWebPart.MetricType.PRECURSOR_AREA).getAttribute("style").contains("background-color: rgb(255, 0, 0)"));
+        Assert.assertTrue("Incorrect heatmap color for lowest(White)",
+                outlierHeatmap.getCellElement(1, QCPlotsWebPart.MetricType.ISOTOPE_DOTP).getAttribute("style").contains("background-color: rgb(255, 255, 255)"));
```
labkey-tchad: Avoid `assertTrue` and `assertFalse` except for very simple assertions. If this assertion fails, it could be a huge pain to figure out what the problem is.
AssertJ has methods that will check for partial text and give a useful error message (e.g. `assertThat(cellStyle).contains("background-color: rgb(255, 0, 0)")`)

You can also pluck out individual style attributes from an element (e.g. `cellElement.getCssValue("background-color")`), which could let you use `assertEquals` instead of checking a substring.

ankurjuneja (approved):

test/src/org/labkey/test/tests/targetedms/TargetedMSOutlierHeatmapTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+        log("Verify Custom date range");
+        outlierHeatmap.setDateRange(OutlierHeatmapSummaryWebPart.HeatmapDateRange.Custom_Range, false);
+        outlierHeatmap.doAndWaitForTableUpdate(() -> {
+            outlierHeatmap.setStartDate("2013-08-01");
+            outlierHeatmap.setEndDate("2013-08-15");
+        });
```
labkey-tchad: I would have a separate method for custom date range. If possible, tests shouldn't need to know when an update is expected.
```suggestion
        outlierHeatmap.setCustomDateRange("2013-08-01", "2013-08-15");
```

test/src/org/labkey/test/components/targetedms/OutlierHeatmapSummaryWebPart.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public class Elements extends BodyWebPart<?>.ElementCache
+    {
+        final Select dateRange = SelectWrapper.Select(Locator.id("date-range")).refindWhenNeeded(this);
+        final Input startDate = new Input(Locator.id("start-date").refindWhenNeeded(this), getDriver());
+        final Input endDate = new Input(Locator.id("end-date").refindWhenNeeded(this), getDriver());
+        final Locator heatmapLoc = Locator.id("heatmap-table");
+        final Table heatmapTable = new Table(getDriver(), heatmapLoc.refindWhenNeeded(this));
+        final WebElement replicateCount = Locator.id("total-replicates").refindWhenNeeded(this);
```
labkey-tchad: All these refinding webelements makes me worried that `doAndWaitForTableUpdate` and `waitForTable` aren't waiting for the right thing. `refindWhenNeeded` should really be used only as a last resort.

labkey-sweta: Updated

test/src/org/labkey/test/tests/targetedms/TargetedMSOutlierHeatmapTest.java
labkey-sweta: Done

labkey-tchad (approved):

PR #943, opened 2024-08-11
Issue 50925: Refinements to peptide outlier heat map

ankurjuneja:
#### Rationale
Following refinements are made in this PR-
1. Show a loading indicator, both on the initial load and after toggling the date range.
2. Show the number of replicates included in the date range instead of the total number in the folder.
3. Get rid of the arrows on the side of the heat map legend.
4. Vertically align the legend text with the center of the legend boxes.
5. Reduce the number of boxes in the legend and add borders to make it consistent with the calendar heat map legend.
6. Show the maximum number of outliers in the legend instead of "Most Outliers", making it consistent with the calendar views.
7. Add the web part between the calendar and the QC summary web parts in targetedms-qcSummaryHistory.view.
8. Re-title the "QC Summary" web part to "Replicate Summary" and "Outlier Heatmap Summary" to "Peptide/Molecule Summary"
9. Sort any non-precursor based rows at the bottom.
10. Show the unmodified peptide sequence. Don't let it wrap between the charge state and the MZ.

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels:
Testing notes. Some are preexisting behaviors, some are new:

1. Preexisting QC folders are missing their QC Summary web parts. We should be backwards compatible with the old name. See BaseWebPartFactory.addLegacyNames().
2. Loading indicator looks good, though it's misleadingly cleared if you select a different date range twice in rapid succession. The indicator clears when the first response comes back, and then the content eventually gets swapped out with the second response. Ideally we'd keep the indicator spinning and perhaps even cancel the first request. Lower priority.
3. When there are zero outliers for the time range, the table cells are blank instead of showing 0. The total columns show 0. If you have a single outlier anywhere in the table, the other cells get zeros. We should show zeros instead of blanks.
4. I'm still seeing "QC Summary" as the web part title instead of "Replicate Summary" (see #8 from the original list)

I imagine those will be small changes, but I'll hold my code review until you've had a chance to push fixes.

labkey-jeckels (changes requested):
See my previously posted comments for some testing feedback.

src/org/labkey/targetedms/TargetedMSServiceImpl.java
```diff
@@ -116,6 +120,43 @@ public List<SampleFileInfo> getSampleFiles(Container container, User user, Integ
...
+    public List<SampleFileInfo> getSampleFiles(Container container, User user, Date startDate, Date endDate)
+    {
+        List<SampleFileInfo> sampleFileInfos = TargetedMSManager.get().getSampleFileInfos(container, user, null);
+        List<SampleFileInfo> result = new ArrayList<>();
+        if (!sampleFileInfos.isEmpty())
+        {
+            for (SampleFileInfo sampleFileInfo : sampleFileInfos)
+            {
+                DateFormat format = DateFormat.getDateInstance();
+                String acquiredTime = format.format(sampleFileInfo.getAcquiredTime());
```
labkey-jeckels: Why do we format this as a string and then immediately parse it back to a Date?

src/org/labkey/targetedms/TargetedMSServiceImpl.java
```diff
@@ -116,6 +120,43 @@ public List<SampleFileInfo> getSampleFiles(Container container, User user, Integ
...
+    @Override
+    public List<SampleFileInfo> getSampleFiles(Container container, User user, Date startDate, Date endDate)
+    {
+        List<SampleFileInfo> sampleFileInfos = TargetedMSManager.get().getSampleFileInfos(container, user, null);
+        List<SampleFileInfo> result = new ArrayList<>();
+        if (!sampleFileInfos.isEmpty())
+        {
+            for (SampleFileInfo sampleFileInfo : sampleFileInfos)
+            {
+                DateFormat format = DateFormat.getDateInstance();
```
labkey-jeckels: We should use DateUtil.parseDate().

labkey-jeckels (approved):
A couple of small requests but no need for further review.

src/org/labkey/targetedms/TargetedMSServiceImpl.java
ankurjuneja: I was doing this to get the date only part. Thanks for the pointer, replaced it with DateUtil.getDateOnly

PR #948, opened 2024-08-15
Issue 50842: Fix upper and lower bounds legends on qc plots

ankurjuneja:
#### Rationale
Hide upper bound and lower bound legends on fixed value cutoff for 'percent of mean' and 'standard deviation' yAxis scale

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels (changes requested):
The change looks OK but I think you need to merge the latest from release24.7-SNAPSHOT into your platform FB to get TC to build and test

labkey-jeckels (approved):

PR #949, opened 2024-08-16
Issue 50995: Highlight replicate on multi-series plot too

ankurjuneja:
#### Rationale
Highlight the replicate for multi series plots when clicked from qc summary metric link.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api-src/org/labkey/api/targetedms/model/SampleFileInfo.java
+++ b/api-src/org/labkey/api/targetedms/model/SampleFileInfo.java
@@ -105,6 +105,7 @@ public class SampleFileInfo extends OutlierCounts
         jsonObject.put("AcquiredTime", getAcquiredTime());
         jsonObject.put("GuideSetId", getGuideSetId());
         jsonObject.put("ReplicateName", getReplicateName());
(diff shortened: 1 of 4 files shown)


labkey-jeckels (changes requested):
I found a number of problems when testing.

I pushed a fix for one - treating the ReplicateId and SampleFileId as the same thing. In some servers they could be equivalent depending on the data files loaded, but they'll typically not match.

Remaining problems:
1. I'm seeing JS errors on the multi-series plot in both Safari and Chrome: Invalid value for <rect> attribute width="NaN" I think it's trying to set values for the single-series plots that aren't being shown.
2. I have a folder locally (/Panorama/Big QC) where the highlighting is off-by-one in the plot (both single and multi-series). Clicking on the last replicate, it doesn't highlight. Clicking on the second-to-last replicate highlights the last one in the plot. I can't figure out why. Let me know if you want to screen share.

labkey-jeckels (approved):
Tested with the latest changes and confirmed that the issues I reported are fixed.

PR #982, opened 2024-11-18
Set qcTraceName on ChromatogramGroupId

vagisha:
#### Rationale
qcTraceName is written to the skyd file but was not being set on ChromatogramGroupId in the Panorama code.
If the name of a QC trace is empty, it cannot be used to create a trace metric in QC folders (https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=50773).  

#### Related Pull Requests
- https://github.com/LabKey/targetedms/pull/931
- https://github.com/LabKey/targetedms/pull/980 (closed because we cannot assume that when there are multiple traces they will be in the same order for all the sample file)

#### Changes
- Set the qcTraceName on ChromatogramGroupId.
- Remove changes added in PR:931 since it creates a unique trace name per sample file instead of using the same name for a trace in all samples. 
- Backported https://github.com/LabKey/targetedms/pull/921


vagisha:
@nickshulman & @josheckels This PR does not set the SpectrumClassFilter on ChromatogramGroupId.  Looks like that was added in this PR: https://github.com/LabKey/targetedms/pull/921, but it was merged with "develop" in July so it is not part of 24.7-SNAPSHOT that we just deployed on PanoramaWeb. 
Should I merge Nick's changes to ChromatogramGroupId and add the new SpectrumFilter class from Nick's PR into this PR?

labkey-jeckels:
@vagisha yes, I think we can back port all of Nick's change to 24.7. We're lucky here that 24.7 already has 24.004 as the schema version for `targetedms` so there's no need to make any of the upgrades conditional.

labkey-jeckels (commented):
This change looks good but I'll wait for the back port to approve

vagisha:
@labkey-jeckels changes have been backported.  Waiting for tests to pass.

labkey-jeckels (approved):

PR #1006, opened 2025-01-30
Optimize chrom import - don't check transition mz if matched on TextId

labkey-jeckels:
#### Rationale
We're spending a lot of time pulling transition chromatogram headers during import on very large documents.

#### Changes
Only do the time consuming work of counting the number of transition matches if more than one chromatogram group was found for a particular Replicate.

### DIFF
--- a/src/org/labkey/targetedms/parser/SkylineBinaryParser.java
+++ b/src/org/labkey/targetedms/parser/SkylineBinaryParser.java
@@ -261,15 +261,12 @@ public class SkylineBinaryParser
         return _cacheFiles != null ? _cacheFiles.length : 0;
     }
 
-    public boolean matchTransitions(ChromGroupHeaderInfo header, List<? extends GeneralTransition> transitions, Double explicitRt, double tolerance)
+    /**
+     * Returns the number of transitions from the Precursor in the Skyline document that have a matching
+     * chromatogram in the ChromGroupHeaderInfo.
+     */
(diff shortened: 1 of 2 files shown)


nickshulman:
I think the signature of "tryLoadChromatogram" needs to be changed so that it takes a SkylineReplicate.
In Skyline, the work of counting the number of transition m/z's which match chromatograms in the chromatogram group is only used to decide which chromatograms to use from a multi-inject replicate.
This comparison is not supposed to be happening between replicates.
If "tryLoadChromatogram" was only looking at the result files from a single replicate (which is usually only one result file) then the optimization would be that it does not need to do the work of matching transition m/z's unless it found more than one candidate chromatogram group.
I will try coding up that change to tryLoadChromatogram.

labkey-jeckels:
> I think the signature of "tryLoadChromatogram" needs to be changed so that it takes a SkylineReplicate. In Skyline, the work of counting the number of transition m/z's which match chromatograms in the chromatogram group is only used to decide which chromatograms to use from a multi-inject replicate. This comparison is not supposed to be happening between replicates. If "tryLoadChromatogram" was only looking at the result files from a single replicate (which is usually only one result file) then the optimization would be that it does not need to do the work of matching transition m/z's unless it found more than one candidate chromatogram group. I will try coding up that change to tryLoadChromatogram.

Thanks Nick. Looks like we might need to move the call `tryLoadChromatogram()` inside of `populateChromInfoChromatograms()`. But feel free to make a change to help keep this analogous to the Skyline code.

nickshulman (approved):

PR #1015, opened 2025-02-19
Issue 52112: Let pressure trace metrics grab the min or max value within a time range

ankurjuneja:
#### Rationale
Let pressure trace metrics grab the first, last, min or max value within a time range

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/resources/schemas/dbscripts/postgresql/targetedms-25.000-25.001.sql
+++ b/resources/schemas/dbscripts/postgresql/targetedms-25.000-25.001.sql
@@ -1,0 +1,3 @@
+ALTER TABLE targetedms.QCMetricConfiguration RENAME COLUMN TimeValue To MinTimeValue;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN MinTimeValue REAL;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN TimeValueOption VARCHAR(10);
--- a/resources/schemas/dbscripts/postgresql/targetedms-25.000-25.001.sql
+++ b/resources/schemas/dbscripts/postgresql/targetedms-25.000-25.001.sql
@@ -1,3 +1,3 @@
 ALTER TABLE targetedms.QCMetricConfiguration RENAME COLUMN TimeValue To MinTimeValue;
-ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN MinTimeValue REAL;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN MaxTimeValue REAL;
 ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN TimeValueOption VARCHAR(10);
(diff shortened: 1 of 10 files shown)


resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
```diff
@@ -108,71 +117,144 @@ Ext4.define('Panorama.Window.AddTraceMetricWindow', {
...
+                                    xtype: 'container',
+                                    layout: 'hbox',
+                                    items: [
+                                        {
+                                            xtype: 'radio',
+                                            name: 'metricValue',
+                                            inputValue: 'timeValue',
+                                            boxLabel: 'Use the',
+                                            width: 70,
+                                            checked: this.operation === this.update ? this.metric.MinTimeValue > 0 : true,
```
labkey-jeckels: Let's allow 0

```suggestion
                                            checked: this.operation === this.update ? this.metric.MinTimeValue >= 0 : true,
```

resources/schemas/dbscripts/postgresql/targetedms-25.000-25.001.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE targetedms.QCMetricConfiguration RENAME COLUMN TimeValue To MinTimeValue;
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN MaxTimeValue REAL;
```
labkey-jeckels, line 2: We need to set values to avoid breaking existing configs.

I think that means something like:

UPDATE targetedms.QCMetricConfiguration SET MaxTimeValue = 1000 WHERE MinTimeValue IS NOT NULL;
UPDATE targetedms.QCMetricConfiguration SET TimeValueOption = 'First' WHERE MinTimeValue IS NOT NULL;

Can you give this a test and commit?

test/src/org/labkey/test/tests/panoramapremium/TargetedMSQCPremiumTest.java
```diff
@@ -193,7 +193,9 @@ public void testTraceMetric()
         String metricName = "Pressure At 5";
         String traceName = "ColumnOven_FC_BridgeFlow (channel 5)";
         String yAxisLabel = "psi";
-        String timeValue = "5";
+        String minTimeValue = "5";
+        String maxTimeValue = "10";
+        String timeValueOption = "First";
```
labkey-jeckels: Let's add a metric for min and max too.

labkey-jeckels (approved):
A couple of requests. Also check out the tweaks I pushed for the dialog text and allowing 0

test/src/org/labkey/test/tests/panoramapremium/TargetedMSQCPremiumTest.java
ankurjuneja: done

resources/schemas/dbscripts/postgresql/targetedms-25.000-25.001.sql
ankurjuneja: tested

PR #1033, opened 2025-03-19
Fix skydDataId on TargetedMSRun

vagisha:
#### Rationale
When a sky.zip file or its exploded folder are moved, post-import, so that the relative locations of sky.zip and its corresponding .skyd file are non-standard, two ExpData rows are created for the skyd file in the Panorama Public copy pipeline job.
- The first ExpData (linked to the ExpRun) is created during XAR import.
- The second ExpData (not linked to the ExpRun) is created in the `SkylineDocumentParser.parseChromatograms()` method.
Normally, while running the copy pipeline job, `SkylineDocumentParser.parseChromatograms()` does not have to create a new ExpData, since an ExpData with the expected path already exists.

Having 2 ExpDatas causes:
- The skydDataId in TargetedMSRun references an ExpData not linked to the ExpRun. It refers to a file in the 'export' directory which gets deleted after folder import. 
- FK violations during cleanup (`CopyExperimentFinalTask.cleanupExportDirectory()`) prevents deletion of ExpData corresponding to the skydDataId

#### Related Pull Requests
- https://github.com/LabKey/MacCossLabModules/pull/526

#### Changes
- Added a method in TargetedMSService to update the skydDataId

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSManager.java
+++ b/src/org/labkey/targetedms/TargetedMSManager.java
@@ -865,10 +865,16 @@ public class TargetedMSManager
 
     public static boolean updateSkydDataId(ITargetedMSRun run, ExpData newSkydData, User user)
     {
-        if (!(run instanceof TargetedMSRun) || newSkydData == null)
+        if (!(run instanceof TargetedMSRun))
         {
-            return false;
+            throw new IllegalArgumentException("Invalid run type. Expected TargetedMSRun but received " +
+                    (run != null ? run.getClass().getName() : "null"));
         }
+        if (newSkydData == null)
+        {
+            throw new IllegalArgumentException("Cannot update with null newSkydData. A valid ExpData object is required");
+        }
+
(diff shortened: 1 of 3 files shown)


src/org/labkey/targetedms/TargetedMSManager.java
```diff
@@ -861,6 +863,24 @@ public static TargetedMSRun getRunByLsid(String lsid, Container c)
         throw new IllegalArgumentException("More than one TargetedMS runs found for LSID "+lsid);
     }
 
+    public static boolean updateSkydDataId(ITargetedMSRun run, ExpData newSkydData, User user)
+    {
+        if (!(run instanceof TargetedMSRun) || newSkydData == null)
+        {
+            return false;
```
labkey-jeckels: Consider making this more aggressive and throw an `IllegalArgumentException` as they seem like a bigger missed assumption than the other conditions below.

labkey-jeckels (approved):

PR #1047, opened 2025-04-02
Fix default values for boolean columns for 24.11->25.3 upgrade

labkey-jeckels:
#### Rationale
Our conditional upgrade code for servers already upgraded to version 25.001 of the schema didn't handle default values for boolean columns correctly.

#### Changes
* Default needs to be a string `'0'`, not an integer `0`


ankurjuneja (approved):

labkey-adam:
@labkey-jeckels are those really double single quotes surrounding the values? Is that intentional?

labkey-adam:
> @labkey-jeckels are those really double single quotes surrounding the values? Is that intentional?

Oh, now that I look at the whole file I see this is all inside a single quoted string. Never mind...

PR #1070, opened 2025-05-03
Improvements to MAM reports and Excel exports

labkey-jeckels:
#### Rationale
Users have requested improvements to the MAM reports. This includes showing more details when a peptide sequence is repeated across protein sequences and showing more cross-linking possibilities. They also want to see richer formatting in Excel exports, especially around modified peptide sequences.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/6627

#### Changes
* Refactor modified peptide rendering to support both Excel and HTML output
* Use rich text formatting to highlight modified peptides in Excel exports
* Overrides to force shared strings in Excel exports, required for rich text support
* Show all indices for peptides repeated across proteins in the same Skyline file in Peptide Map
* Smaller formatting updates for Peptide Map report
* Make better use of HtmlString and DOM builder

### DIFF
--- a/test/src/org/labkey/test/tests/targetedms/TargetedMSPeptideLibraryTest.java
+++ b/test/src/org/labkey/test/tests/targetedms/TargetedMSPeptideLibraryTest.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.test.tests.targetedms;
 
+import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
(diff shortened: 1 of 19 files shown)


test/src/org/labkey/test/tests/targetedms/TargetedMSPeptideLibraryTest.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.labkey.test.tests.targetedms;
 
+import org.apache.commons.lang3.StringUtils;
```
ankurjuneja: unused?

ankurjuneja (approved):

test/src/org/labkey/test/tests/targetedms/TargetedMSPeptideLibraryTest.java
labkey-jeckels: Yes. Removed.

PR #1077, opened 2025-05-15
Migrate to renderGridCellContents(RenderContext, HtmlWriter)

labkey-adam:
#### Rationale
Server-side HTML generation wants to use DOM, builders, and HtmlWriter


labkey-jeckels (approved):
Did light testing of the FB and all looked good.

PR #1089, opened 2025-06-26
Improved test coverage for peptide heatmap report

labkey-jeckels:
#### Rationale
We've done a number of bug fixes for the peptide heat map summary report and want to avoid regressions.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/2517

#### Changes
* Check more types of date ranges, including error conditions
* Make sure custom dates are sticky
* Assorted spot checks of peptides and metrics

### DIFF
--- a/test/src/org/labkey/test/components/targetedms/PeptideSummaryWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/PeptideSummaryWebPart.java
@@ -38,20 +38,51 @@ public class PeptideSummaryWebPart extends BodyWebPart<PeptideSummaryWebPart.Ele
 
     public PeptideSummaryWebPart setStartDate(String value)
     {
-        elementCache().startDate.set(value);
+        if (value != null)
+        {
+            elementCache().startDate.set(value);
+        }
+        else
+        {
+            elementCache().startDate.clear();
+        }
         return this;
     }
 
     public PeptideSummaryWebPart setEndDate(String value)
     {
-        elementCache().endDate.set(value);
+        if (value != null)
+        {
+            elementCache().endDate.set(value);
+        }
+        else
+        {
+            elementCache().endDate.clear();
+        }
         return this;
     }
 
-    public PeptideSummaryWebPart apply()
+    public String getStartDate()
     {
-        doAndWaitForElementToRefresh(() -> elementCache().applyButton.findElement(this).click(),
-                elementCache().heatmapLoc, getWrapper().defaultWaitForPage);
+        return elementCache().startDate.get();
+    }
+
+    public String getEndDate()
+    {
+        return elementCache().endDate.get();
+    }
+
+    public PeptideSummaryWebPart apply(boolean wait)
+    {
+        if (wait)
+        {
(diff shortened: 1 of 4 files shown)


test/src/org/labkey/test/components/targetedms/PeptideSummaryWebPart.java
```diff
@@ -38,20 +38,51 @@ public PeptideSummaryWebPart setDateRange(HeatmapDateRange value)
...
+    {
+        if (wait)
+        {
+            doAndWaitForElementToRefresh(() -> elementCache().applyButton.findElement(this).click(),
+                    elementCache().heatmapLoc, getWrapper().defaultWaitForPage);
+        }
+        else
+        {
+            elementCache().applyButton.findElement(this).click();
+        }
```
labkey-danield: Looking at the test code that passes in false, and assuming I understand the intent, generally the design pattern is to have a second method named `applyExpectingError`. 
The `applyExpectingError` would click then wait for the error to show up. What it does after clicking does vary a bit. It could assert that an error is shown, take a parameter (of some kind) and assert the error matches, return the error (message, element, dialog, etc...) and let the test decide what to do.

Example:
```
    public PeptideSummaryWebPart apply()
    {
        doAndWaitForElementToRefresh(() -> elementCache().applyButton.findElement(this).click(),
                elementCache().heatmapLoc, getWrapper().defaultWaitForPage);
        return this;
    }


    public PeptideSummaryWebPart applyExpectingError(String errorMessage)
    {
        elementCache().applyButton.findElement(this).click();
        getWrapper().waitForText(errorMessage);
    }

```
Of course, if I don't understand the intent then... nevermind.

test/src/org/labkey/test/tests/targetedms/TargetedMSPeptideSummaryHeatmapTest.java
```diff
@@ -61,13 +64,59 @@ public void testHeatMapColorAndValues()
...
+        waitAndClickAndWait(Locator.linkContainingText("View all 47 replicates"));
+        // Make sure the custom date range matches the other plot's
+        peptideSummaryHeatMap = new PeptideSummaryWebPart(getDriver());
+        assertEquals("2013-08-10", peptideSummaryHeatMap.getStartDate());
+        assertEquals("2013-08-15", peptideSummaryHeatMap.getEndDate());
+
+        log("Verify invalid date combos produce helpful errors");
+        peptideSummaryHeatMap.setCustomDateRange("2013-08-15", "2013-08-01");
+        peptideSummaryHeatMap.apply(false);
+        assertTextPresent("Please choose a start date that is before the end date.");
```
labkey-danield: For example, this would become:

`peptideSummaryHeatMap.applyExpectingError("Please choose a start date that is before the end date.");`

labkey-danield (approved):
Your change will work, but consider the change I suggested for the `PeptideSummaryWebPart.apply` method.

test/src/org/labkey/test/components/targetedms/PeptideSummaryWebPart.java
labkey-jeckels: Sounds good. Updated to use that convention.

PR #1104, opened 2025-08-11
Fix for Update button not visible when chromatograms chart display settings form is expanded

vagisha:
#### Rationale
The chromatograms chart display settings form had a fixed height of 160px and was rendered into a hidden container, causing the Update button to be cut off when the form content exceeded this height.

#### Changes
- Replaced fixed height (160px) with autoHeight for proper sizing.
- Create form when "Display Chart Settings" is expanded (not on page load) to ensure proper rendering in a visible container.


labkey-jeckels (approved):
I tested this locally and it worked well.

PR #1105, opened 2025-08-11
Issue 53623: Improve QC metrics config UI

labkey-jeckels:
#### Rationale
We can simplify the setup for QC metrics, make it possible to edit or delete dual-series configurations, and prevent errors from being as fatal.

#### Changes
* Assume all metric queries are in the `targetedms` schema
* Tolerate series without a label
* Don't let a bad query prevent imports of new documents
* Allow editing of metrics that define two different series

- [x] Automated tests
- [x] Manual testing @labkey-tchad

### DIFF
--- a/resources/views/configureQCMetric.html
+++ b/resources/views/configureQCMetric.html
@@ -11,9 +11,9 @@
         LABKEY.internal = {};
 
     LABKEY.internal.ConfigureQCMetrics = new function () {
-        var qcMetrics;
-        var qcMetricsTable ='';
-        var configRows = [];
+        let qcMetrics;
+        let qcMetricsTable ='';
+        let configRows = [];
 
         function showQCMetrics() {
             qcMetricsTable += '<form id="qcMetricsForm" >';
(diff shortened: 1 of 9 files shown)


vagisha (approved):
I tested this locally and was able to create a two-series metric without having to select schemas, and I was also able to edit the metric. Thanks! 
A couple of observations: 
- I am not sure what the **Series 1 Axis Label** and **Series 2 Axis Label** fields are for. They are not being used to label the Y-axes. 
- For the **Transition & Precursor Area** metric, one of the Y-axis label colors does not  match the corresponding series color in the plot.

labkey-tchad:
Manual test results:
 - Axis label is still required for first series
 - Axis labels don't appear in QC plots
 - Dashboard and QC Summary History pages are broken if any bad custom metrics are defined
   - Repro: Create a custom QC metric that references a non-existent query
   - `Error: Error on line 4: Query or table not found: targetedms.noexist`
 - Replicate Summary grid on QC Summary History page doesn't include include custom metrics that have axis labels defined for both series (it does include the first series if on it has a label).
 - CalibrationCurve, MoleculeCalibrationCurve, and PeptiteCalibrationCurve all show up as "CalibrationCurve" in the custom metric dialog's query selection dropdown. This might be more of problem with `GetQueriesAction` but I wanted to make not of it while I continue to investigate.

labkey-jeckels:
Thanks @labkey-tchad and @vagisha. You are correct that the labeling is messed up, and the dual-series metrics are problematic. I'd like to merge this change as-is (after my correction for the triple `CalibrationCurve` in the query list) and tackle the rest separately.

labkey-jeckels:
> Thanks @labkey-tchad and @vagisha. You are correct that the labeling is messed up, and the dual-series metrics are problematic. I'd like to merge this change as-is (after my correction for the triple `CalibrationCurve` in the query list) and tackle the rest separately.

I should add that these are existing issues, not introduced by this PR.

PR #1114, opened 2025-09-22
TargetedMS Test Fixess

labkey-danield:
#### Rationale
Some minor clean up.
Trying to address the failure in [PassportTest.testSteps](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_External_Panorama_PanoramaBPostgres/3676087?buildTab=tests&status=failed&name=PassportTest.testSteps). This may be related to the current flakiness with external API calls to EBI (example: [https://www.ebi.ac.uk/proteins/api/features/P02774](https://www.ebi.ac.uk/proteins/api/features/P02774) returns a 500), but waiting for the elements may not be a bad thing.

#### Related Pull Requests
* None

#### Changes
* Use waitAndClickAndWait.
* Use waitForElement(s).

### DIFF
--- a/test/src/org/labkey/test/tests/targetedms/TargetedMSLightHeavyRatioTest.java
+++ b/test/src/org/labkey/test/tests/targetedms/TargetedMSLightHeavyRatioTest.java
@@ -47,7 +47,7 @@ public class TargetedMSLightHeavyRatioTest extends TargetedMSTest
         clickTab("Proteins");
         DataRegionTable table = new DataRegionTable("PeptideGroup", getDriver());
         table.setFilter("Label", "Equals", "gi|324021745|ref|NP_001191236.1|");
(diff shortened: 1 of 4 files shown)


labkey-tchad (commented):
I don't see any reason to increase all of these timeouts. I don't think the failures we're seeing are going to be prevented by longer waits. An external server is just having issues.

labkey-danield:
> I don't see any reason to increase all of these timeouts. I don't think the failures we're seeing are going to be prevented by longer waits. An external server is just having issues.

Yeah, I've reverted most of the timeout changes. The only changes now are a few locators and using waitAndClickAndWait.

If you still think these changes are not necessary I can close the PR.

labkey-tchad (approved):

PR #1116, opened 2025-09-30
Instrument scheduling improvements

labkey-jeckels:
[Original spec](https://docs.google.com/document/d/1WBEy6crd8k92u2pLMkZUY7BYjH8HcscLiqKM29MBUy4/edit?tab=t.0)

#### Rationale
Incorporate feedback on the initial instrument scheduling work to get it closer to production usage.

#### Changes
* Add monthly billing report
* Add view that shows schedules for all instruments at once
* Add webpart to help admins navigate setup, `Instrument Scheduling Admin`
* Stop ending schedule events one minute short of the hour
* Layout, formatting, and minor usability improvements

- [x] Manual Testing @labkey-danield  📍 
- [x] Automated test @labkey-jeckels
- [x] Bug fixes @labkey-jeckels 📍 
- [x] Code review


resources/queries/targetedms/instrumentBilling.query.xml
```diff
@@ -0,0 +1,16 @@
+<query xmlns="http://labkey.org/data/xml/query">
+    <metadata>
+        <tables xmlns="http://labkey.org/data/xml">
+            <table tableName="instrumentSchedule" tableDbType="TABLE">
```
ankurjuneja:
```suggestion
            <table tableName="instrumentBilling" tableDbType="TABLE">
```

resources/queries/targetedms/instrumentBillingByMonth.query.xml
```diff
@@ -0,0 +1,16 @@
+<query xmlns="http://labkey.org/data/xml/query">
+    <metadata>
+        <tables xmlns="http://labkey.org/data/xml">
+            <table tableName="instrumentSchedule" tableDbType="TABLE">
```
ankurjuneja:
```suggestion
            <table tableName="instrumentBillingByMonth" tableDbType="TABLE">
```

webapp/TargetedMS/js/scheduleUtils.js
```diff
@@ -0,0 +1,58 @@
...
+        const r = parseInt(c.substring(0, 2), 16);
+        const g = parseInt(c.substring(2, 4), 16);
+        const b = parseInt(c.substring(4, 6), 16);
+        const yiq = ((r * 299) + (g * 587) + (b * 114)) / 1000;
+        return yiq >= 128 ? '#000000' : '#FFFFFF';
+    };
+
+    // Return a time-only format string without seconds/millis based on the container setting
+    utils.getTimeOnlyFormat = function() {
+        if (window.LABKEY && LABKEY.container && LABKEY.container.formats && LABKEY.container.formats.timeFormat) {
```
ankurjuneja:
```suggestion
        if (window.LABKEY && LABKEY.container?.formats?.timeFormat) {
```

ankurjuneja (approved):

labkey-jeckels:
@labkey-danield I've got a basic automated test in place. I'm reluctant to do more until we have more feedback from the lab. For example, they've mentioned they're not sure they want to continue supporting splitting payment across methods for a single reservation.

vagisha:
Sorry for the delay in reviewing. I pulled the branch on my dev machine, and the multi-instrument calendar overview looks great. Thank you for adding this view. 

Here are some potential issues I noticed, along with a few suggestions for future improvement:

**Pre-existing bugs (not introduced in this PR)**
- Clicking **Add Payment Method** link (**Project Details** page - `targetedms-msProjectDetails.view`)
  - The project is not pre-selected in the **Project** dropdown.
  - Clicking Submit or Cancel results in "_No project selected_" error because the `returnUrl `is missing the `project` param.
  - Likely cause: the URL is missing the `project` param in `msProjectDetails.html`. It should look something like:
```
const urlParams = {
            schemaName: 'targetedms',
            queryName: 'ProjectPaymentMethod',
            project: LABKEY.ActionURL.getParameter('project'),
            returnUrl: LABKEY.ActionURL.buildURL('targetedms', 'msProjectDetails', null, { project:  LABKEY.ActionURL.getParameter('project')})
        };
```
- **Payment Method filtering**
When scheduling instrument time, the **Payment Method** dropdown shows all payment methods (not just those linked to the project). Ideally, only project-linked methods should appear.
(_Note: This is how MSDaPl is set up, but the MacCoss lab may prefer not to require linking anymore. Worth clarifying with Genn and Rich_.)
- **Duplicate project-payment method rows**
It’s currently possible to link the same project and payment method more than once, which creates duplicate rows in `projectPaymentMethod `table

**Suggestions**
- **Project Details page**
  - Add a link to add a researcher to the project.
  - Show an overview of instrument time already scheduled for the project.
- **Your Projects** grid
  - Is it possible to add a button to **Add Project** directly from here?
- The **Instrument Scheduling Admin** webpart is very helpful.  I think we should only need to use it for 
  - Adding Instrument, Rate Type, Instrument Rate
  - Viewing the instrument billing report
Other things like project / researcher mapping, adding payment method, project / payment method mapping should all be possible to do through the project details page.

vagisha:
One more comment. In the **Your Projects** grid:
- All projects are listed, even for users who are not researchers on those projects.
- There’s also an error above the grid: "_Ignoring filter/sort on column 'researcher' because it does not exist_."
This is because `userProjects.sql` doesn’t include the `researcher` column, but `msProject.html` is trying to filter on it.

image

 Currently, `msProject.html` applies the `researcher` filter only for non-admins.  But I think that admins like Genn and Rich would probably prefer to see only their projects in the **Your Projects** grid by default. There could be a button in the grid that lets admins view all projects.

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
```diff
@@ -0,0 +1,221 @@
...
+import org.labkey.test.util.PostgresOnlyTest;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.List;
+import java.util.Map;
+
+@Category({})
+@BaseWebDriverTest.ClassTimeout(minutes = 5)
```
labkey-danield: I don't think you need this ClassTimeout annotation.

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
```diff
@@ -0,0 +1,221 @@
...
+        yearMonth += month;
+
+        scheduleInstrument(yearMonth + "-02");
+        scheduleInstrument(yearMonth + "-03");
+        scheduleInstrument(yearMonth + "-03", true);
+        scheduleInstrument(yearMonth + "-03");
+
+        assertProjectEventCounts(2, 0);
+
+        sleep(1000);  // Wait for the dialog to clear out of the way
```
labkey-danield: I'm not going to object to using a sleep. If you want something that is more dependable on the dialog going away you could try waiting for one of the elements in the dialog to become stale:
        `shortWait().until(ExpectedConditions.stalenessOf(Locator.byClass("activeProjectEvent").findElement(getDriver())));
`

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
```diff
@@ -0,0 +1,221 @@
...
+        scheduleInstrument(yearMonth + "-05");
+        assertProjectEventCounts(2, 2);
+
+        sleep(1000);  // Wait for the dialog to clear out of the way
+        doAndWaitForPageToLoad(() -> selectOptionByText(Locator.id("instrumentDropDown"), INSTRUMENT_2));
+        scheduleInstrument(yearMonth + "-06");
+        assertProjectEventCounts(1, 0);
+
+        goToDashboard();
+        clickAndWait(Locator.linkWithText("All instrument calendar view"));
```
labkey-danield: In cases like this, you navigate and then immediately click on an element on the page, I've started using `waitAndClickAndWait`. It will wait a little bit for the element to show up before it tries to click it.

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
```diff
@@ -0,0 +1,221 @@
...
+        // Schedule for hours within a day instead of 24-hour periods
+        // Check billing for individual months, including reservations that span month boundaries with start/end dates
+        // Ensure that overlapping reservations are rejected
+        // Ensure that reservations cannot be made for inactive instruments
+    }
+
+    private void assertProjectEventCounts(int expectedActiveCount, int expectedOtherCount)
+    {
+        assertElementPresent(Locator.tagWithClass("div", "activeProjectEvent"), expectedActiveCount);
+        assertElementPresent(Locator.tagWithClass("div", "otherProjectEvent"), expectedOtherCount);
```
labkey-danield: Alan recently made the argument that we should be using `Locator.byClass("activeProjectEvent")` for robustness. I don't think this project needs to worry about that, but wanted to raise it.

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
labkey-jeckels: In this case it's being rendered as HTML in the server code, so I don't think the initial wait is needed. But I can follow that pattern if it doesn't slow things down

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
labkey-jeckels: Sure, I can avoid using the tag though I am not anticipating changing this to be a `<h2>` or something

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
labkey-jeckels: Removed

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
labkey-jeckels: I switched to something similar

labkey-danield:
A couple of more issues:

1. **Instrument Filter is not Applied**
Go to the All Instruments View (targetedms-scheduleAllInstruments.view)
Set Project to All Projects.
Filter to one instrument.
Change project filter to a specific project.
Result:
The instrument filter is no longer applied. It looks like all instruments for that project are shown.

2. **When scheduling and instrumenty the value in Instrument drop down resets if project is changed.**
From a project (say Project 1) click schedule an instrument.
In the Schedule Instrument Time page (/targetedms-scheduleInstrument.view):
Select a different instrument from the default instrument.
Change the operator.
Change Project to a different project.
Result:
The value in the Instrument field resets to the default instrument.

I think this behavior with the value in the Instrument drop down is how I "appear" to get overlapping schedules...

3. **The value in the Instrument drop down is applied if I edit an existing schedule.**
From a project click schedule an instrument.
In the Schedule Instrument Time page (/targetedms-scheduleInstrument.view):
Select a different instrument from the default instrument.
Save the schedule
From the same project again schedule an instrument.
This time select an existing schedule and edit it (change it's time)
Result:
A new schedule was created for the "default" instrument that is selected in the Instrument drop down.

labkey-jeckels:
> A couple of more issues:
> 
> 1. **Instrument Filter is not Applied**
>    Go to the All Instruments View (targetedms-scheduleAllInstruments.view)
>    Set Project to All Projects.
>    Filter to one instrument.
>    Change project filter to a specific project.
>    Result:
>    The instrument filter is no longer applied. It looks like all instruments for that project are shown.
> 2. **When scheduling and instrumenty the value in Instrument drop down resets if project is changed.**
>    From a project (say Project 1) click schedule an instrument.
>    In the Schedule Instrument Time page (/targetedms-scheduleInstrument.view):
>    Select a different instrument from the default instrument.
>    Change the operator.
>    Change Project to a different project.
>    Result:
>    The value in the Instrument field resets to the default instrument.
> 
> I think this behavior with the value in the Instrument drop down is how I "appear" to get overlapping schedules...
> 
> 3. **The value in the Instrument drop down is applied if I edit an existing schedule.**
>    From a project click schedule an instrument.
>    In the Schedule Instrument Time page (/targetedms-scheduleInstrument.view):
>    Select a different instrument from the default instrument.
>    Save the schedule
>    From the same project again schedule an instrument.
>    This time select an existing schedule and edit it (change it's time)
>    Result:
>    A new schedule was created for the "default" instrument that is selected in the Instrument drop down.

These should be fixed. I'm less confident about the third but I'm unable to reproduce it.

labkey-danield:
> > A couple of more issues:
> > 
> > 1. **Instrument Filter is not Applied**
> >    Go to the All Instruments View (targetedms-scheduleAllInstruments.view)
> >    Set Project to All Projects.
> >    Filter to one instrument.
> >    Change project filter to a specific project.
> >    Result:
> >    The instrument filter is no longer applied. It looks like all instruments for that project are shown.
> > 2. **When scheduling and instrumenty the value in Instrument drop down resets if project is changed.**
> >    From a project (say Project 1) click schedule an instrument.
> >    In the Schedule Instrument Time page (/targetedms-scheduleInstrument.view):
> >    Select a different instrument from the default instrument.
> >    Change the operator.
> >    Change Project to a different project.
> >    Result:
> >    The value in the Instrument field resets to the default instrument.
> > 
> > I think this behavior with the value in the Instrument drop down is how I "appear" to get overlapping schedules...
> > 
> > 3. **The value in the Instrument drop down is applied if I edit an existing schedule.**
> >    From a project click schedule an instrument.
> >    In the Schedule Instrument Time page (/targetedms-scheduleInstrument.view):
> >    Select a different instrument from the default instrument.
> >    Save the schedule
> >    From the same project again schedule an instrument.
> >    This time select an existing schedule and edit it (change it's time)
> >    Result:
> >    A new schedule was created for the "default" instrument that is selected in the Instrument drop down.
> 
> These should be fixed. I'm less confident about the third but I'm unable to reproduce it.

Yep, these are fixed. I don't see any other issues that would prevent merging this work.

PR #1120, opened 2025-10-06
Improve QC folder plot controls

labkey-jeckels:
#### Rationale
We've accumulated a lot of controls for Panorama's QC plots. We can do a better job of arranging them and explaining them.

#### Changes
* Group related controls together in three columns
* Eliminate Apply button for date ranges
* Do better at throwing previous requests when users rapidly update settings
* Move some error messaging out of dialogs
* Switch from checkboxes to radio buttons to better explain behavior

### DIFF
--- a/webapp/TargetedMS/js/QCPlotHelperBase.js
+++ b/webapp/TargetedMS/js/QCPlotHelperBase.js
@@ -84,7 +84,7 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
     getPlotsData: function() {
         // get input number N
         // pass includeTrailingCV or includeTrailingMean in plotsConfig
-        var plotsConfig = {};
+        const plotsConfig = {};
         plotsConfig.metricId = this.metric;
         plotsConfig.metricId2 = this.metric2;
         plotsConfig.includeLJ = this.showMetricValuePlot();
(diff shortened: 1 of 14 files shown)


labkey-tchad (approved):
Test updates look good.

labkey-jeckels:
@cnathe could you give the UI changes a code review?

webapp/TargetedMS/js/QCPlotHelperBase.js
```diff
@@ -121,14 +121,44 @@ Ext4.define("LABKEY.targetedms.QCPlotHelperBase", {
...
+
+        const failureCb = LABKEY.Utils.getCallbackWrapper(this.failureHandler, this);
+
+        this._qcActiveRequest = LABKEY.Ajax.request({
             url: LABKEY.ActionURL.buildURL('targetedms', 'GetQCPlotsData.api'),
             success: function(response) {
-                this.lastParsedResponse = JSON.parse(response.responseText);
-                this.processPlotData();
+                // Ignore if not the most recent request
+                if (requestSeq !== this._qcRequestSeq)
```
cnathe, line 139: since we are calling `this._qcActiveRequest.abort()` before creating a new request, it seems like this case of "not the most recent request" shouldn't be hit. Is this just for the rare case where we don't abort in time?

cnathe (approved):

webapp/TargetedMS/js/QCPlotHelperBase.js
labkey-jeckels: Yes. It might be overkill for this scenario, but what I saw was that abort() can still be subject to a race condition. I would have thought that JS's default threading model would make this a non-issue, but decided to do both since it wasn't much extra work

PR #1129, opened 2025-10-16
Cache precursor-scoped metrics for better reporting performance

labkey-jeckels:
#### Rationale
On very large databases, the QC plots and other reports can be sluggish to load, especially on parent folders that have multiple QC subfolders.

We can improve perf by caching all of the precursor-scoped metric values in a dedicated table. We then maintain it as Skyline documents are added or removed, or metrics are added or removed.

#### Changes
* New `targetedms.QCMetricCache` table
* Populate lazily at upgrade and then on imports/deletes or metric configuration changes
* Button to manually recache after editing a custom metric's query
* Eliminate "enabled" queries as they're no longer necessary to speed things up


resources/queries/targetedms/QCMetric_peakArea.sql
```diff
@@ -17,4 +17,40 @@ SELECT
...
\ No newline at end of file
+FROM PrecursorChromInfo
+WHERE
+    -- Ensure we have precursor areas for peptides or small molecules
+    (EXISTS (SELECT Id, FragmentType, Quantitative FROM Transition t WHERE FragmentType = 'precursor' AND Charge IS NULL)
+        OR EXISTS (SELECT Id, FragmentType, Quantitative FROM MoleculeTransition t WHERE FragmentType = 'precursor' AND Charge IS NULL))
+
+  -- Ensure we have transition areas for peptides or small molecules
+AND (EXISTS (
+    SELECT Id, FragmentType, Quantitative FROM Transition t
```
ankurjuneja: Could the two EXISTS subqueries being used multiple times be moved into a separate sql file and also, would using a JOIN than two EXISTS improve perf here?

ankurjuneja (approved):

resources/queries/targetedms/QCMetric_peakArea.sql
labkey-jeckels: Good idea. I restored the transition and precursor area enabled queries and used them here.

PR #1154, opened 2025-12-24
Handle creating a duplicate custom new metric

ankurjuneja:
#### Rationale
https://github.com/LabKey/internal-issues/issues/591

#### Related Pull Requests
* 

#### Changes
* display error message when custom metric with same name exists

### DIFF
--- a/resources/web/PanoramaPremium/window/AddNewMetricWindow.js
+++ b/resources/web/PanoramaPremium/window/AddNewMetricWindow.js
@@ -285,36 +285,82 @@ Ext4.define('Panorama.Window.AddCustomMetricWindow', {
         this.queryError.setVisible(!isValid);
     },
 
+    checkMetricNameExists: function (metricName, callback) {
+        let filterArray = [LABKEY.Filter.create('Name', metricName, LABKEY.Filter.Types.EQUAL)];
+
+        // If updating, exclude the current metric from the check
+        if (this.operation === this.update && this.metric) {
+            filterArray.push(LABKEY.Filter.create('id', this.metric.id, LABKEY.Filter.Types.NOT_EQUAL));
+        }
+
+        LABKEY.Query.selectRows({
+            containerPath: LABKEY.container.id,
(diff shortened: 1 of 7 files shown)


resources/web/PanoramaPremium/window/AddNewMetricWindow.js
```diff
@@ -285,36 +285,82 @@ Ext4.define('Panorama.Window.AddCustomMetricWindow', {
...
-                method: 'POST',
-                success: function () {
-                    window.location.reload();
+        if(isValid) {
+            let metricName = this.metricNameField.getValue();
+
+            // Check if metric name already exists
+            this.checkMetricNameExists(metricName, function (exists) {
+                if (exists) {
+                    let errorMessage = 'A metric with the name "' + Ext4.util.Format.htmlEncode(metricName) + '" already exists. Please choose a different name.';
```
labkey-jeckels: Since you're using `setText()`, this ends up double-encoding.,

labkey-jeckels (changes requested):
Please add some simple automated test coverage for this scenario.

In manual testing, I confirmed that this fixed the custom metric error handling. There's a similar repro for trace metrics that we should also address here.

test/src/org/labkey/test/tests/panoramapremium/TargetedMSQCFolderImportExport.java
```diff
@@ -118,7 +118,7 @@ private void createGuideSet(String projectName)
         Assert.assertEquals("Guide Set was not added correctly", 1, table.getDataRowCount());
     }
 
-    private void addCustomMetric(String projectName, String metricName, String queryName)
+    private void addCustomMetric(String projectName, String metricName, String queryName, boolean checkForDuplicate)
```
labkey-jeckels: Would `duplicateNameErrorExpected` be a better name for this parameter?

labkey-jeckels (approved):

PR #1163, opened 2026-01-12
Handle timezone correctly during saving qc annotation

ankurjuneja:
#### Related Pull Requests
* 

#### Changes
*


ankurjuneja:
@labkey-tchad  The test seems to be fixed now in different time zones. @labkey-jeckels should I fix this in 26.1 or here develop is OK?

labkey-jeckels:
> @labkey-tchad The test seems to be fixed now in different time zones. @labkey-jeckels should I fix this in 26.1 or here develop is OK?

Develop is fine.

labkey-jeckels (approved):

PR #1169, opened 2026-01-23
Require a login to protect public folders against aggressive bots

labkey-jeckels:
#### Rationale
The transition list is an expensive page to render for very large Skyline documents. A bot has been hitting public pages on PanoramaWeb over 28,000 times in 24 hours, causing perf problems.

#### Changes
* Require a login to view the transition list, even when guests have read access to the container


vagisha (approved):

labkey-jeckels:
Looks good. I tested as a guest and saw the login prompt on those three actions. I also tested that requesting Show All or a count higher than 50 only returned 50.

PR #1170, opened 2026-01-27
Fix malformed links on the Passport Details page

vagisha:
#### Rationale
Chromatogram and peptide detail links were malformed due to a missing `?` before URL query parameters.

#### Changes
- Use LABKEY.ActionURL.buildURL to build the links


labkey-jeckels (approved):
We should add an automated test. Doesn't have to be this PR. I assume this broke a while ago when stopped always adding `?` to the end of the URL.

PR #1171, opened 2026-01-28
Make TargetedMSModule unavailable when not active in container

labkey-jeckels:
#### Rationale
Modules like TargetedMS don't work very well when they're not active in a container. Specifically, their schema isn't available, so many actions (export, plotting, etc) break.

#### Changes
* Check if active (or we're in the root)


labkey-jeckels:
@vagisha FYI. I don't expect any problems on PanoramaWeb as everything is using the folder type, which will have the module active. If someone tries to hit an action in `TargetedMSController` where the module isn't active, there will be an error page that says the module isn't available.

labkey-tchad (approved):

PR #1172, opened 2026-01-28
Replace hospscotch tooltip JS library for QC Plots with tippy

ankurjuneja:
#### Rationale
https://github.com/LabKey/internal-issues/issues/787

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/7367

#### Changes
* change qc plots data point tooltip from hopscotch to tippy
* transform qcplothoverpanel from ext js implementation to plain js implementation

### DIFF
--- a/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
@@ -530,10 +530,11 @@ public final class QCPlotsWebPart extends BodyWebPart<QCPlotsWebPart.Elements>
                     WebElement point = getPointByAcquiredDate(acquiredDate);
                     ScrollUtils.scrollIntoView(point, center, center);
                     getWrapper().mouseOverWithoutScrolling(point);
-                    return getWrapper().isElementPresent(Locator.tagWithClass("div", "x4-form-display-field")
-                            .containing(acquiredDate.substring(0, 16))); // drop seconds part (e.g. "2013-08-12 04:54") for trailing mean/CV
+                    return getWrapper().isElementPresent(Locator.tagWithClass("div", "qc-plot-hover-panel")
+                            .withDescendant(Locator.tagWithClass("div", "qc-hover-field")
+                                    .containing(acquiredDate.substring(0, 16)))); // drop seconds part (e.g. "2013-08-12 04:54") for trailing mean/CV
                 });
-        return elementCache().hopscotchBubble.findElement(getDriver());
+        return elementCache().tippyBubble.findElement(getDriver());
     }
 
     @LogMethod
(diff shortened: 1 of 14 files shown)


labkey-jeckels (commented):
Can we remove hopscotch as a dependency from this module? It looks like qcTrendPlotReport.jsp still references it. qcSummary.jsp does too. The latter looks like it's still actively using it. There are other references to hopscotch in the codebase - could they be replaced (they may just be CSS style names, etc)

ankurjuneja:
> Can we remove hopscotch as a dependency from this module? It looks like qcTrendPlotReport.jsp still references it. qcSummary.jsp does too. The latter looks like it's still actively using it. There are other references to hopscotch in the codebase - could they be replaced (they may just be CSS style names, etc)

Will do, I just replaced the qc plots data points tooltip, will replace other usage too.

test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
```diff
@@ -708,37 +692,24 @@ public void checkAllPlotTypes(boolean selected)
...
     {
-        return Locator.byClass("hopscotch-bubble-container");
+        return Locator.tagWithClass("div", "qc-plot-hover-panel");
     }
 
     public Locator.XPathLocator getBubbleContent()
     {
-        Locator.XPathLocator hopscotchBubble = Locator.byClass("hopscotch-bubble-container");
-        return hopscotchBubble.append(Locator.byClass("hopscotch-bubble-content").append(Locator.byClass("hopscotch-content").withText()));
+        return Locator.tagWithClass("div", "qc-plot-hover-panel");
```
labkey-jeckels: Seems redundant with `getBubble()` now.

test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
```diff
@@ -970,8 +941,7 @@ public class Elements extends BodyWebPart<?>.ElementCache
         Locator.CssLocator paginationPrevBtn = Locator.css(".qc-paging-prev");
         Locator.CssLocator paginationNextBtn = Locator.css(".qc-paging-next");
         Locator.CssLocator svgBackgrounds = Locator.css("svg g.brush rect.background");
-        Locator.XPathLocator hopscotchBubble = Locator.byClass("hopscotch-bubble-container");
-        Locator.XPathLocator hopscotchBubbleClose = Locator.byClass("hopscotch-bubble-close");
+        Locator.XPathLocator tippyBubble = Locator.tagWithClass("div", "qc-plot-hover-panel");
```
labkey-jeckels, line 939: Should this use `getBubble()`? Or should that use this?

webapp/TargetedMS/js/QCPlotHoverPanel.js
```diff
@@ -77,251 +90,350 @@ Ext4.define('LABKEY.targetedms.QCPlotHoverPanel', {
...
+        labelSpan.style.width = '120px';
+        labelSpan.style.fontWeight = 'bold';
+        labelSpan.style.flexShrink = '0';
+        labelSpan.textContent = label + ':';
+
+        var valueSpan = document.createElement('span');
+        valueSpan.className = 'qc-hover-field-value';
+        valueSpan.style.flex = '1';
+        valueSpan.style.wordBreak = 'break-all';
+        valueSpan.innerHTML = value;
```
labkey-jeckels: I don't see where HTML encoding is happening. Should this be `innerText`, or is the caller responsible for encoding (and actually doing it)?

labkey-jeckels (approved):
Code is looking good, though see comments.

We should get manual testing on this before merging. I haven't had a chance to do that yet.

PR #1173, opened 2026-01-29
Add link to "Register" for a new account

vagisha:
#### Rationale
We will enable self-sign-up on PanoramaWeb.  For pages that don't allow guest access, add a link to "Register" for an account in case the user does not have one.
image

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/1169

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSController.java
+++ b/src/org/labkey/targetedms/TargetedMSController.java
@@ -133,6 +133,7 @@ import org.labkey.api.reports.model.ViewCategoryManager;
 import org.labkey.api.reports.report.RedirectReport;
 import org.labkey.api.reports.report.ReportDescriptor;
 import org.labkey.api.security.ActionNames;
+import org.labkey.api.security.AuthenticationManager;
 import org.labkey.api.security.Group;
 import org.labkey.api.security.LoginUrls;
 import org.labkey.api.security.RequiresLogin;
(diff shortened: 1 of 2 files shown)


src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -4461,13 +4461,20 @@ public String getDataRegionNameSmallMolecule()
...
         return new HtmlView(DOM.createHtmlFragment(
                 DOM.DIV(cl("alert alert-info"),
-                        "Please ",
                         DOM.A(at(style, "font-weight: bold;",
                                  href, PageFlowUtil.urlProvider(LoginUrls.class).getLoginURL(container, context.getActionURL())),
-                                "login"),
-                        " to view this data")));
+                                "Login"),
+                        " to view this data.",
+                       DOM.BR(),
```
labkey-jeckels: I suggest conditionalizing this with a check to AuthenticationManager.isRegistrationEnabled(). Otherwise, on other servers it would send the user to a 404.

src/org/labkey/targetedms/TargetedMSController.java
```diff
@@ -4461,13 +4461,20 @@ public String getDataRegionNameSmallMolecule()
     @NotNull
     private static HtmlView getLoginView(ViewContext context, Container container)
     {
+        ActionURL registerUrl = new ActionURL("login", "register", container);
```
labkey-jeckels: I opened a PR to expose this via `LoginUrls`. Please switch to use that.

https://github.com/LabKey/platform/pull/7372

labkey-jeckels (changes requested):
A couple of small requests.

Seems like we should add a basic test for this. We don't need to cover the self-registration case, but we could add to an existing test. Give guests access to the folder, logout, check that you can see a page or two, and that you get this prompt when you hit a blocked action.

labkey-jeckels (approved):
Looks good!

PR #1183, opened 2026-03-06
Share instrument annotations across QC folders

ankurjuneja:
#### Rationale
Panorama users want to record changes to the system (MP, SSTs, columns, PMs, etc) via the Annotations features and having them in separate folders requires them to add the Annotations to both even though it is the same system. 
This will simplify the need by adding the shared annotations for the same instrument across multiple QC folders.

#### Related Pull Requests
* 

#### Changes
* Add new columns to QCAnnotation and QCAnnotationType tables
* Associate shared annotations with instrument of the QC folder
* Display shared annotations in qc folders with same instruments across server

### DIFF
--- a/resources/queries/targetedms/qcannotation/.qview.xml
+++ b/resources/queries/targetedms/qcannotation/.qview.xml
@@ -4,6 +4,8 @@
         <column name="Date" />
         <column name="EndDate" />
         <column name="QCAnnotationTypeId" />
+        <column name="instrumentModel" />
+        <column name="instrumentSerialNumber" />
         <column name="Container" />
         <column name="Created" />
         <column name="Modified" />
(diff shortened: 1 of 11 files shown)


ankurjuneja:
@labkey-jeckels should this go in 26.3 or develop as it has a sql upgrade script?

labkey-jeckels:
> @labkey-jeckels should this go in 26.3 or develop as it has a sql upgrade script?

26.3 please. We get ourselves in trouble when we have upgrade scripts in multiple active branches.

I'll wait to review until you've had a chance to update the target branch.

labkey-jeckels:
@ankurjuneja we can delete this PR and its branch at this point, correct?

PR #1184, opened 2026-03-10
Ensure FKs and indices in targetedms schema

ankurjuneja:
#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSModule.java
+++ b/src/org/labkey/targetedms/TargetedMSModule.java
@@ -231,7 +231,7 @@ public class TargetedMSModule extends SpringModule implements ProteomicsModule
     @Override
     public Double getSchemaVersion()
     {
-        return 26.002;
+        return 26.004;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/org/labkey/targetedms/TargetedMSModule.java
```diff
@@ -231,7 +231,7 @@ public String getName()
     @Override
     public Double getSchemaVersion()
     {
-        return 26.002;
+        return 26.004;
```
labkey-adam, line 234: It doesn't really matter, but I'm curious why 26.003 was skipped.

labkey-adam:
@ankurjuneja TargetedMS development (especially new SQL scripts) is typically merged to the current release branch, so 26.3 currently. Maybe there's a reason to target develop for this PR, but please double check: once an upgrade script is merged to develop no further scripts can be added to 26.3.

src/org/labkey/targetedms/TargetedMSModule.java
ankurjuneja: it is in another PR in the works- https://github.com/LabKey/targetedms/pull/1183/changes

PR #1185, opened 2026-03-19
Share instrument annotations across QC folders

ankurjuneja:
#### Rationale
Panorama users want to record changes to the system (MP, SSTs, columns, PMs, etc) via the Annotations features and having them in separate folders requires them to add the Annotations to both even though it is the same system. 
This will simplify the need by adding the shared annotations for the same instrument across multiple QC folders.

#### Related Pull Requests
* 

#### Changes
* Add new columns to QCAnnotation and QCAnnotationType tables
* Associate shared annotations with instrument of the QC folder
* Display shared annotations in qc folders with same instruments across server

### DIFF
--- a/resources/schemas/dbscripts/postgresql/targetedms-26.002-26.003.sql
+++ b/resources/schemas/dbscripts/postgresql/targetedms-26.002-26.003.sql
@@ -1,5 +1,5 @@
 ALTER TABLE targetedms.QCAnnotationType ADD COLUMN Shareable BOOLEAN DEFAULT FALSE;
-ALTER TABLE targetedms.QCAnnotation ADD COLUMN instrumentModel VARCHAR(255);
-ALTER TABLE targetedms.QCAnnotation ADD COLUMN instrumentSerialNumber VARCHAR(255);
+ALTER TABLE targetedms.QCAnnotation ADD COLUMN instrumentModel VARCHAR(300);
+ALTER TABLE targetedms.QCAnnotation ADD COLUMN instrumentSerialNumber VARCHAR(200);
 
 UPDATE targetedms.QCAnnotationType SET Shareable = TRUE WHERE Name = 'Instrumentation Change';
--- a/resources/schemas/dbscripts/postgresql/targetedms-26.002-26.003.sql
+++ b/resources/schemas/dbscripts/postgresql/targetedms-26.002-26.003.sql
@@ -1,0 +1,5 @@
+ALTER TABLE targetedms.QCAnnotationType ADD COLUMN Shareable BOOLEAN DEFAULT FALSE;
+ALTER TABLE targetedms.QCAnnotation ADD COLUMN instrumentModel VARCHAR(255);
+ALTER TABLE targetedms.QCAnnotation ADD COLUMN instrumentSerialNumber VARCHAR(255);
+
+UPDATE targetedms.QCAnnotationType SET Shareable = TRUE WHERE Name = 'Instrumentation Change';
--- a/src/org/labkey/targetedms/TargetedMSManager.java
+++ b/src/org/labkey/targetedms/TargetedMSManager.java
@@ -17,6 +17,8 @@
 package org.labkey.targetedms;
 
 import com.google.common.base.Joiner;
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
@@ -207,7 +209,8 @@ public class TargetedMSManager
 
     public static List<SampleFileChromInfo> getSampleFileChromInfos(SampleFile sampleFile)
     {
-        return new TableSelector(getTableInfoSampleFileChromInfo(), new SimpleFilter(FieldKey.fromParts("SampleFileId"), sampleFile.getId()), new Sort("TextId")).getArrayList(SampleFileChromInfo.class);    }
+        return new TableSelector(getTableInfoSampleFileChromInfo(), new SimpleFilter(FieldKey.fromParts("SampleFileId"), sampleFile.getId()), new Sort("TextId")).getArrayList(SampleFileChromInfo.class);
+    }
 
     public static SampleFileChromInfo getSampleFileChromInfo(int id, Container c)
     {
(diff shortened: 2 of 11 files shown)


resources/schemas/dbscripts/postgresql/targetedms-26.002-26.003.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE targetedms.QCAnnotationType ADD COLUMN Shareable BOOLEAN DEFAULT FALSE;
+ALTER TABLE targetedms.QCAnnotation ADD COLUMN instrumentModel VARCHAR(255);
+ALTER TABLE targetedms.QCAnnotation ADD COLUMN instrumentSerialNumber VARCHAR(255);
```
labkey-jeckels: `targetedms.instrument.model` is 300 characters. `targetedms.samplefile.instrumentserialnumber` is 200. We should match

src/org/labkey/targetedms/TargetedMSManager.java
```diff
@@ -3046,6 +3054,41 @@ private QueryUpdateService getNicknameUpdateService(User user, Container contain
...
+        private Long instrumentId;
+
+        public InstrumentDetails()
+        {
+        }
+    }
+
+    public static List<InstrumentDetails> getInstrumentDetails(Container container)
+    {
+        SQLFragment sql = new SQLFragment("SELECT DISTINCT sf.InstrumentSerialNumber, i.Model, i.Id AS InstrumentId FROM ");
```
labkey-jeckels: See testing notes. Do we need the Id here? It's causing problems.

test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
```diff
@@ -959,6 +960,58 @@ public void testSamplesExcludedForSingleMetric()
...
+                        annotation.getDescription().equals(shareableAnnotation.getDescription()))
+                {
+                    shareableAnnotationFound = true;
+                    break;
+                }
+            }
+            if (shareableAnnotationFound)
+                break;
+        }
+        assertTrue("Shareable annotation from Folder A should appear in Folder B QC plots", shareableAnnotationFound);
```
labkey-jeckels: Also test that the user can't see annotations in folders they don't have read access to. I manually tested and confirmed that it worked correctly.

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1301,68 +1301,168 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
+                        let separator = "";
+                        for (let i = 0; i < instrumentData.rows.length; i++) {
+                            let row = instrumentData.rows[i];
+                            let model = row["Model"];
+                            let serial = row["InstrumentSerialNumber"];
+
+                            instrumentFilter += separator + "(";
+                            let innerSep = "";
+                            if (model) {
+                                instrumentFilter += "(qca.instrumentModel = '" + model + "'";
```
labkey-jeckels: Not a true security risk since LabKey SQL protects us, but we should escape the string here to deal with quotes.

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1301,68 +1301,168 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
+                                innerSep = " AND ";
+                            } else {
+                                instrumentFilter += "(qca.instrumentModel IS NULL";
+                                innerSep = " AND ";
+                            }
+
+                            if (serial) {
+                                instrumentFilter += innerSep + "qca.instrumentSerialNumber = '" + serial + "')";
```
labkey-jeckels: Escape the string

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1909,23 +2009,54 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
+        let mouseOn = function(pt, strokeWidth, d) {
             d3.select(pt).transition().duration(800).attr("stroke-width", strokeWidth).ease("elastic");
+
+            if (!pt._tippy) {
+                let date = new Date(d['Date']);
+                let dateStr = me.formatDate(date, date.getHours() !== 0 || date.getMinutes() !== 0 || date.getSeconds() !== 0);
+                let content = "<table>"
+                        + "<tr><td style='vertical-align: top; padding-right: 5px;'>Created By:</td><td>" + LABKEY.Utils.encodeHtml(d['DisplayName']) + "</td></tr>"
+                        + "<tr><td style='vertical-align: top; padding-right: 5px;'>Type:</td><td>" + LABKEY.Utils.encodeHtml(d['Name']) + "</td></tr>"
+                        + "<tr><td style='vertical-align: top; padding-right: 5px;'>Date:</td><td>" + dateStr + "</td></tr>"
```
labkey-jeckels: Should be safe without it, but let's encode this value for completeness.

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -2018,15 +2163,48 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
+                        const record = combo.getStore().findRecord('Id', newValue);
+                        const isShared = record ? record.get('Shareable') : false;
+                        const field = combo.up('window').down('#shared-annotation-display');
+                        field.setVisible(isShared);
+                    }
+                }
+            }, {
+                xtype: 'displayfield',
+                itemId: 'shared-annotation-display',
+                value: '<span style="color: #555;"><i class="fa fa-share-alt"></i> Shared</span>',
```
labkey-jeckels:
```suggestion
                value: '<span style="color: #555;"><i class="fa fa-share-alt"></i> Shared with other folders using this instrument</span>',
```

or something like that

labkey-jeckels (commented):
Looking good! A few small requests. Will retest when you've responded to the manual test feedback and approve then.

webapp/TargetedMS/js/QCTrendPlotPanel.js
ankurjuneja: I tried this earlier, It adds all the escape strings into the value displayed

labkey-jeckels: It works OK for me, including when I hack the code to append `'<>'` to the date string. What's it looking like for you? Hacked version to ensure we're not double-encoding:

image

ankurjuneja: seems to be resolving now, I promise it was showing all kinds of escape characters when I saw earlier, anyway I added the encoding.

labkey-jeckels (approved):
Looks good. Test failures but they may be unrelated to this branch.

PR #1187, opened 2026-03-25
Ensure FKs and indices in targetedms schema

ankurjuneja:
#### Rationale
Add missing foreign key to core.Containers and add missing dedicated index on containers on columns.
26.002-26.003 upgrade script is here https://github.com/LabKey/targetedms/pull/1185

#### Related Pull Requests
* 

#### Changes
These tables have the Container column but lack a formal FOREIGN KEY constraint to the core.Containers table.
* targetedms.Runs 
* targetedms.QCAnnotation
* targetedms.GuideSet
* targetedms.AutoQCPing
* targetedms.PrecursorChromInfo
* targetedms.SampleFileChromInfo

missing index:

* targetedms.QCAnnotationType 
* targetedms.QCAnnotation 
* targetedms.GuideSet 
* targetedms.QCMetricConfiguration

### DIFF
--- a/resources/schemas/dbscripts/postgresql/targetedms-26.003-26.004.sql
+++ b/resources/schemas/dbscripts/postgresql/targetedms-26.003-26.004.sql
@@ -1,5 +1,12 @@
--- Adding missing foreign keys to core.Containers(EntityId)
 SELECT core.executeJavaUpgradeCode('reparentOrphanedTargetedMSData');
+
+-- Adding missing foreign keys to core.Containers(EntityId)1
+ALTER TABLE targetedms.Runs ADD CONSTRAINT FK_Runs_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
+ALTER TABLE targetedms.QCAnnotation ADD CONSTRAINT FK_QCAnnotation_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
+ALTER TABLE targetedms.GuideSet ADD CONSTRAINT FK_GuideSet_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
+ALTER TABLE targetedms.AutoQCPing ADD CONSTRAINT FK_AutoQCPing_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
+ALTER TABLE targetedms.PrecursorChromInfo ADD CONSTRAINT FK_PrecursorChromInfo_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
+ALTER TABLE targetedms.SampleFileChromInfo ADD CONSTRAINT FK_SampleFileChromInfo_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
 
 -- Adding missing indices on Container
 CREATE INDEX IX_QCAnnotationType_Container ON targetedms.QCAnnotationType(Container);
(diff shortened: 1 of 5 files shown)


ankurjuneja:
The failing test is already assigned in TC.

resources/schemas/dbscripts/postgresql/targetedms-26.003-26.004.sql
```diff
@@ -0,0 +1,13 @@
+-- Adding missing foreign keys to core.Containers(EntityId)
+ALTER TABLE targetedms.Runs ADD CONSTRAINT FK_Runs_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
+ALTER TABLE targetedms.QCAnnotation ADD CONSTRAINT FK_QCAnnotation_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
+ALTER TABLE targetedms.GuideSet ADD CONSTRAINT FK_GuideSet_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
+ALTER TABLE targetedms.AutoQCPing ADD CONSTRAINT FK_AutoQCPing_Container FOREIGN KEY (Container) REFERENCES core.Containers(EntityId);
```
labkey-jeckels, line 7: These three could easily be deleted as part of the upgrade in case there are orphaned rows.

The other three would be harder because they have many more FKs pointed at them. Maybe we should do what I helped the MacCoss lab do - reparent any orphaned rows in Runs, PrecursorChromInfo, or SampleFileChromInfo to the /Shared project before doing the delete.

src/org/labkey/targetedms/TargetedMSListener.java
```diff
@@ -71,5 +71,6 @@ public void containerDeleted(Container c, User user)
         new SqlExecutor(TargetedMSManager.getSchema()).execute("DELETE FROM " + TargetedMSManager.getTableInfoMSInstrument() + " WHERE Container = ?", c);
         new SqlExecutor(TargetedMSManager.getSchema()).execute("DELETE FROM " + TargetedMSManager.getTableInfoPaymentMethod() + " WHERE Container = ?", c);
         new SqlExecutor(TargetedMSManager.getSchema()).execute("DELETE FROM " + TargetedMSManager.getTableInfoRateType() + " WHERE Container = ?", c);
+        new SqlExecutor(TargetedMSManager.getSchema()).execute("DELETE FROM " + TargetedMSManager.getTableInfoSampleFileChromInfo() + " WHERE Container = ?", c);
```
labkey-jeckels: If this is needed, it means that the call to `deleteSampleFileDependent(getTableInfoSampleFileChromInfo())` in `TargetedMSManager.purgeDeletedRuns()` isn't doing what it should, or there's another codepath that should be doing the same thing.

Were there test failures or other repros that inspired adding it?

labkey-jeckels (changes requested):
One or two requests.

src/org/labkey/targetedms/TargetedMSListener.java
ankurjuneja: Yes, there was this test failure https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_External_Panorama_PanoramaBPostgres?branch=Issue894_ensureFK&buildTypeTab=overview#3885366

labkey-jeckels: That's actually exposed a bug in moving runs. It should update the SampleFileChromInfo's container column, like we're doing here for PrecursorChromInfo.

https://github.com/LabKey/targetedms/blob/fbaa1c4b622ae1523bec2452ecd668104462757e/src/org/labkey/targetedms/TargetedMSManager.java#L2508

ankurjuneja:
@labkey-jeckels  can you please review again to see if the changes are capturing the orphaned rows during the upgrade correctly and moving run bug?

labkey-jeckels (approved):

PR #1192, opened 2026-04-04
fix verifyGuideSetRelatedElementsForPlots

ankurjuneja:
#### Rationale
Add cleanup in TargetedMSQCTest as shared annotations were getting displayed in other test containers causing failures. The failures are intermittent because of the ordering of tests execution in TC.

#### Related Pull Requests
* 

#### Changes
*


test/src/org/labkey/test/tests/targetedms/TargetedMSQCTest.java
```diff
@@ -197,6 +199,14 @@ public void preTest()
         qcPlotsWebPart.revertToDefaultView();
     }
 
+    @Override
+    protected void doCleanup(boolean afterTest) throws TestTimeoutException
+    {
+        // Use the API-based approach for deletion so that we don't trigger AJAX requests navigating to the delete page
+        // that may run in the background and cause SQL Server deadlock exceptions
```
labkey-jeckels: Panorama doesn't support SQL Server anymore so this comment can be updated or trimmed.

labkey-jeckels (approved):

PR #1193, opened 2026-04-12
Fix accessibility issues suggested by WAVE

ankurjuneja:
#### Rationale
https://github.com/LabKey/internal-issues/issues/878

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1980
* https://github.com/LabKey/platform/pull/7573

#### Changes
*

### DIFF
--- a/src/org/labkey/targetedms/view/precursorConflictResolutionView.jsp
+++ b/src/org/labkey/targetedms/view/precursorConflictResolutionView.jsp
@@ -271,7 +271,7 @@ function toggleCheckboxSelection(element)
              <!--<td class="representative newPrecursor <%=precursor.getNewPrecursorId()%>"><%=precursor.getNewPrecursorId()%></td>-->
              <td class="representative details-control newPrecursor <%=precursor.getNewPrecursorId()%>">
                 <span class="<%=precursor.getNewPrecursorId()%>_<%=precursor.getOldPrecursorId()%>">
-                     <img src="<%=getWebappURL("_images/plus.gif")%>"/>
+                     <img src="<%=getWebappURL("_images/plus.gif")%>" alt="Expand row details"/>
                  </span>
              </td>
              <td class="representative newPrecursor <%=precursor.getNewPrecursorId()%>">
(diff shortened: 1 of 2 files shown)


src/org/labkey/targetedms/view/precursorConflictResolutionView.jsp
```diff
@@ -271,7 +271,7 @@ function toggleCheckboxSelection(element)
              <!--<td class="representative newPrecursor <%=precursor.getNewPrecursorId()%>"><%=precursor.getNewPrecursorId()%></td>-->
              <td class="representative details-control newPrecursor <%=precursor.getNewPrecursorId()%>">
                 <span class="<%=precursor.getNewPrecursorId()%>_<%=precursor.getOldPrecursorId()%>">
-                     <img src="<%=getWebappURL("_images/plus.gif")%>"/>
+                     <img src="<%=getWebappURL("_images/plus.gif")%>" alt="Expand row details"/>
```
labkey-jeckels, line 274: These should flip when we swap the plus for a minus on lines 122 and 127.

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -1272,10 +1272,10 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
         var btnHtml = '';
 
         btnHtml += '<span class="qc-paging-prev ' + (this.pagingStartIndex > 0 ? 'qc-paging-icon-enabled' : 'qc-paging-icon-disabled')
-                + '"><i class="fa fa-angle-left"></i></span>';
+                + '" role="button" aria-label="Previous page"><i class="fa fa-angle-left" aria-hidden="true"></i></span>';
```
labkey-jeckels, line 1275: <span> isn't keyboard focusable by default. It's probably best to switch these to <button>, which would let us drop the `role="button"` and would mean that we wouldn't need to toggle an `aria-disabled` attribute when the control is disabled.

labkey-jeckels (changes requested):

labkey-jeckels:
@ankurjuneja reminder about this pending PR

PR #1209, opened 2026-05-15
Click/drag zooming for QC plots

ankurjuneja:
#### Rationale
https://github.com/LabKey/internal-issues/issues/1047

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/7683

#### Changes
* QCTrendPlotPanel.js
  - Added yZoomByPlot: {} class property — transient JS-only state, never persisted
  - displayTrendPlot(preserveZoom) — clears yZoomByPlot on every normal re-render (metric change, date change, etc.), but skips the clear when called with preserveZoom=true from the zoom methods
  - getYZoomDomain(plotId) — returns [yMin, yMax] or null
  - applyYZoom(plotId, yMin, yMax) — stores zoom and calls displayTrendPlot(true) to re-render with the new domain
  - resetYZoom(plotId) — deletes the entry and re-renders; includes a TODO for the server metric tracking action

* QCPlotHelperBase.js
  - Both addEachIndividualPrecursorPlot and addEachCombinedPrecursorPlot now check for an active zoom domain and pass it as trendLineProps.yZoomDomain before plot construction
  - Both call this.addYZoomInteraction(plot, id) after plot.render()
  - New addYZoomInteraction(plot, plotId) method: appends a transparent rect over the left margin (y-axis area) with `zoom-in` cursor and a D3 v3 drag behavior. Drag creates a teal selection box; on drag end,
  "Zoom" and "Cancel" SVG buttons appear at the midpoint. "Zoom" calls applyYZoom; "Cancel" clears. When zoom is already active, a "Reset Zoom" SVG text link is rendered just above the plot area

### DIFF
--- a/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
+++ b/test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
@@ -904,6 +904,42 @@ public final class QCPlotsWebPart extends BodyWebPart<QCPlotsWebPart.Elements>
         }
     }
 
+    public void performYAxisZoom(QCPlot qcPlot)
+    {
+        WebElement plotEl = qcPlot.getPlot();
+        WebElement overlay = elementCache().yZoomOverlay.findElement(plotEl);
+        getWrapper().scrollIntoView(overlay);
+
+        int dragOffset = 50;
+        new Actions(getWrapper().getDriver())
+                .moveToElement(overlay, 0, -dragOffset)
+                .clickAndHold()
+                .moveToElement(overlay, 0, dragOffset)
+                .release()
+                .perform();
+
+        WebDriverWrapper.waitFor(() -> !elementCache().yZoomButtons.findElements(plotEl).isEmpty(),
+                "Zoom buttons did not appear after y-axis drag", WAIT_FOR_JAVASCRIPT);
+
+        WebElement buttonsGroup = elementCache().yZoomButtons.findElement(plotEl);
+        Locator.css("g").findElement(buttonsGroup).click();
+    }
+
+    public boolean isZoomActive(QCPlot qcPlot)
+    {
+        return !elementCache().yZoomBorder.findElements(qcPlot.getPlot()).isEmpty();
+    }
+
+    public boolean isResetZoomVisible(QCPlot qcPlot)
+    {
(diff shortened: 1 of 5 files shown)


test/src/org/labkey/test/components/targetedms/QCPlotsWebPart.java
```diff
@@ -904,6 +904,42 @@ public String toString()
...
+                .clickAndHold()
+                .moveToElement(overlay, 0, dragOffset)
+                .release()
+                .perform();
+
+        WebDriverWrapper.waitFor(() -> !elementCache().yZoomButtons.findElements(plotEl).isEmpty(),
+                "Zoom buttons did not appear after y-axis drag", WAIT_FOR_JAVASCRIPT);
+
+        WebElement buttonsGroup = elementCache().yZoomButtons.findElement(plotEl);
+        Locator.css("g").findElement(buttonsGroup).click();
```
labkey-jeckels: We could make this a little more robust by identifying the Zoom button via a class or its text.

webapp/TargetedMS/js/QCTrendPlotPanel.js
```diff
@@ -2223,6 +2226,26 @@ Ext4.define('LABKEY.targetedms.QCTrendPlotPanel', {
...
+        return this.yZoomByPlot && this.yZoomByPlot[plotId] ? this.yZoomByPlot[plotId] : null;
+    },
+
+    applyYZoom: function(plotId, yMin, yMax) {
+        if (!this.yZoomByPlot) {
+            this.yZoomByPlot = {};
+        }
+        this.yZoomByPlot[plotId] = [yMin, yMax];
+        this.displayTrendPlot(true /* preserveZoom */);
+        // TODO: add server-side metric tracking action (targetedms/trackQCPlotAction.api)
```
labkey-jeckels: You can use the existing server-side mechanism. This is TypeScript so you can't call it directly, but it's a small enough chunk to copy. https://github.com/LabKey/labkey-ui-components/blob/6ae3bc6de9954f8480fcb24e828611870f515f06/packages/components/src/internal/actions.ts#L576

labkey-jeckels (approved):
One optional further improvement to consider: if the user does the second click inside of the plot instead of on the axis, it would be nice if it finished setting the zoom range too.

PR #1210, opened 2026-05-17
Easy metrics based on replicate and precursor annotation values

ankurjuneja:
#### Rationale
https://github.com/LabKey/internal-issues/issues/1046

#### Related Pull Requests
* 

#### Changes
*  Add AnnotationName column to QCMetricConfiguration to store the backing annotation name for annotation-based metrics
* Extend OutlierGenerator SQL generation to join ReplicateAnnotation (run-scoped) or PrecursorChromInfoAnnotation (precursor-scoped) when a metric has an AnnotationName.
* Add plain JS/jQuery dialog (AddNewAnnotationMetricWindow.js) for creating and editing annotation-backed metrics.


resources/schemas/dbscripts/postgresql/targetedms-26.007-26.008.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN AnnotationName VARCHAR(200);
```
labkey-jeckels: We should match other tables.
```suggestion
ALTER TABLE targetedms.QCMetricConfiguration ADD COLUMN AnnotationName VARCHAR(255);
```

resources/views/configureQCMetric.html
```diff
@@ -80,7 +83,10 @@
                 LABKEY.internal.ConfigureQCMetrics.addNewMetric('custom');
             });
             jQuery('#createNewTraceMetricButton').click(function() {
-                LABKEY.internal.ConfigureQCMetrics.addNewMetric('trace')
+                LABKEY.internal.ConfigureQCMetrics.addNewMetric('trace');
```
labkey-jeckels: Use a constant here and other places, including for `annotation`

resources/web/PanoramaPremium/window/AddNewAnnotationMetricWindow.js
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+
+    function checkMetricNameExists(metricName, callback) {
+        const filterArray = [LABKEY.Filter.create('Name', metricName, LABKEY.Filter.Types.EQUAL)];
+        if (_config.operation === 'update' && _config.metric) {
+            filterArray.push(LABKEY.Filter.create('id', _config.metric.id, LABKEY.Filter.Types.NOT_EQUAL));
+        }
+        LABKEY.Query.selectRows({
+            containerPath: LABKEY.container.id,
```
labkey-jeckels: While this should work in practice, better to use the path.

```suggestion
            containerPath: LABKEY.container.path,
```

src/org/labkey/targetedms/outliers/OutlierGenerator.java
```diff
@@ -100,6 +100,30 @@ private String getEachSeriesTypePlotDataSql(QCMetricConfiguration configuration)
             sql.append(" WHERE metric = ").append(configuration.getId());
             sql.append(")");
         }
+        else if (configuration.getAnnotationName() != null)
+        {
+            // annotation-backed metrics: escape the annotation name for SQL string literal
+            String escapedName = configuration.getAnnotationName().replace("'", "''");
```
labkey-jeckels, line 106: Let's use the metric name instead of the annotation name.

resources/web/PanoramaPremium/window/AddNewAnnotationMetricWindow.js
```diff
@@ -0,0 +1,226 @@
...
+        if (!isValid) {
+            showError('Please fill in all required fields.');
+        }
+        return isValid;
+    }
+
+    function getAnnotationTarget() {
+        return $('input[name="annotationType"]:checked').val() === 'precursor'
+            ? 'precursor_result'
+            : 'replicate';
```
labkey-jeckels, line 58: In the custom query metric dialog and in the table of metrics we use "run" instead of "replicate". I think "replicate" is better. Can you update the other dialog and table to be consistent with the Precursor and Replicate?

resources/web/PanoramaPremium/window/AddNewAnnotationMetricWindow.js
```diff
@@ -0,0 +1,226 @@
...
+        });
+    }
+
+    function buildDialogHtml() {
+        const op = _config.operation;
+        const metric = _config.metric || {};
+        const isPrecursor = op === 'update' && metric.PrecursorScoped;
+        const title = op === 'insert' ? 'Add Annotation-Backed Metric' : 'Edit Annotation-Backed Metric';
+
+        return '<div id="' + DIALOG_ID + '" style="position:fixed;top:0;left:0;width:100%;height:100%;background:rgba(0,0,0,0.5);z-index:9999;display:flex;align-items:center;justify-content:center;">'
```
labkey-jeckels, line 164: This is somewhat frustratingly inconsistent with the query based custom metric dialog, which has almost identical controls. The fields are in different orders, we use a radio vs a drop-down for replicate vs precursor, and the button layout is different.

While I'd like to refactor all of this into React soon, can you adapt the query metric dialog to use this plain HTML/JS approach too?

labkey-jeckels (changes requested):
A few minor suggestions and a larger request to make one of the other metric dialogs consistent with this new approach.

resources/web/PanoramaPremium/window/AddNewAnnotationMetricWindow.js
ankurjuneja: refactored

resources/web/PanoramaPremium/window/AddNewAnnotationMetricWindow.js
ankurjuneja: done

src/org/labkey/targetedms/outliers/OutlierGenerator.java
ankurjuneja: using metricName as label

resources/web/PanoramaPremium/window/AddNewAnnotationMetricWindow.js
```diff
@@ -0,0 +1,226 @@
...
+                }
+            });
+        });
+    }
+
+    function deleteMetric() {
+        if (!confirm('This will delete the "' + _config.metric.name + '" metric. Are you sure?')) return;
+
+        LABKEY.Query.saveRows({
+            containerPath: LABKEY.container.id,
```
labkey-jeckels:
```suggestion
            containerPath: LABKEY.container.path,
```

labkey-jeckels (approved):

PR #1220, opened 2026-05-29
Fix accessibility issues suggested by WAVE

ankurjuneja:
#### Rationale
https://github.com/LabKey/internal-issues/issues/878

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1980
* https://github.com/LabKey/platform/pull/7573

#### Changes
*

### DIFF
--- a/src/org/labkey/targetedms/view/precursorConflictResolutionView.jsp
+++ b/src/org/labkey/targetedms/view/precursorConflictResolutionView.jsp
@@ -119,12 +119,12 @@ $(document).ready(function () {
         {
             row.child.hide();
             tr.removeClass('shown');
-            $("." + cls).children('img').attr('src', "<%=getWebappURL("_images/minus.gif")%>");
+            $("." + cls).children('img').attr('src', "<%=getWebappURL("_images/plus.gif")%>").attr('alt', 'Expand row details');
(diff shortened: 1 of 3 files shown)


labkey-jeckels (changes requested):
See comment

src/org/labkey/targetedms/view/precursorConflictResolutionView.jsp
```diff
@@ -119,12 +119,12 @@ $(document).ready(function () {
         {
             row.child.hide();
             tr.removeClass('shown');
-            $("." + cls).children('img').attr('src', "<%=getWebappURL("_images/plus.gif")%>");
+            $("." + cls).children('img').attr('src', "<%=getWebappURL("_images/minus.gif")%>");
```
labkey-jeckels: Isn't this backwards now? My comment in the previous PR was meant to say that we should update the text when the code toggles between plus and minus, not that we should flip the plus and minus handling.

For example, on line 274 use plus.gif and use expand as the alt text, and do the same on line 122 (instead of just swapping the icon but leaving the original text)

labkey-jeckels (approved):

PR #1222, opened 2026-06-02
Disable LabKey caching of ResultSet by default

labkey-jeckels:
#### Rationale
As we work to reduce OutOfMemoryError risk, we need to have places that rely on caching Results behavior opt-in. Here, it's so that we can ask the size of the result set before iterating it.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/7711

#### Changes
* Opt-in to a `CachedResultSet`


labkey-jeckels:
@labkey-adam this is somewhat more involved. I found more QueryService.select() overloads that should be migrated

labkey-adam (approved):

PR #1231, opened 2026-06-10
Reduce log noise when a spectrum library file is missing or unreadable

vagisha:
#### Rationale
After deploying #1226 on PWEB-DR, the code that hides spectrum libraries from guests based on library file size logged a `NoSuchFileException` stack trace at WARN on every missing .elib/.blib file (PWEB-DR storage is not in sync with PWEB), filling up `labkey.log` very quickly. 

#### Related Pull Requests
- https://github.com/LabKey/targetedms/pull/1226

#### Changes
- Log a single WARN line (with the exception type) instead of the full trace.
- Keep the full stack trace at DEBUG (off by default at INFO) for when it is needed.

### DIFF
--- a/src/org/labkey/targetedms/view/spectrum/LibrarySpectrumMatchGetter.java
+++ b/src/org/labkey/targetedms/view/spectrum/LibrarySpectrumMatchGetter.java
@@ -114,9 +114,10 @@ public class LibrarySpectrumMatchGetter
         }
         catch (IOException e)
         {
-            // If we cannot stat the file it is missing or unreadable, in which case the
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
This seems like a fine improvement.

Depending on the frequency of this logging, you might consider using our simple Throttle utility that can coalesce output in cases like this:

https://github.com/LabKey/platform/blob/0bbf9af4e575c674355dc3b7e0bca218569ce2a5/api/src/org/labkey/api/cache/Throttle.java#L40

PR #1233, opened 2026-06-14
Fix PassportTest and improve scoping

labkey-jeckels:
#### Rationale
PassportTest needs an update to expect the updated data being pulled dynamically from UniProt

#### Changes
* Be tolerant of changes in protein metadata from third-party service
* Add container checks
* Backport query name quoting

### DIFF
--- a/src/org/labkey/targetedms/TargetedMSContainerScopingTest.java
+++ b/src/org/labkey/targetedms/TargetedMSContainerScopingTest.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2026 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.labkey.targetedms;
+
+import jakarta.servlet.http.HttpServletResponse;
+import org.apache.logging.log4j.Logger;
+import org.junit.After;
+import org.junit.Test;
+import org.labkey.api.data.Container;
+import org.labkey.api.data.SimpleFilter;
+import org.labkey.api.data.Table;
+import org.labkey.api.data.TableSelector;
+import org.labkey.api.exp.api.ExpRun;
+import org.labkey.api.exp.api.ExperimentService;
+import org.labkey.api.pipeline.PipeRoot;
+import org.labkey.api.pipeline.PipelineService;
+import org.labkey.api.query.FieldKey;
+import org.labkey.api.security.permissions.AbstractContainerScopingTest;
+import org.labkey.api.util.GUID;
+import org.labkey.api.util.logging.LogHelper;
+import org.labkey.api.view.ActionURL;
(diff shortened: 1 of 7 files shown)


src/org/labkey/targetedms/TargetedMSContainerScopingTest.java
```diff
@@ -0,0 +1,141 @@
...
+import org.labkey.targetedms.parser.skyaudit.UnitTestUtil;
+
+import java.io.File;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Container-scoping integration tests for {@link TargetedMSController} actions that resolve an object by a
+ * global id. Each test sets up the same object in one folder and confirms the action rejects a request that addresses
```
vagisha, line 46: Comment truncated?  Should it be "...that addresses the object through a different folder."?

src/org/labkey/targetedms/TargetedMSContainerScopingTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * ShowSkylineAuditLogExtraInfoAJAXAction renders the full Skyline audit detail (user, timestamp, extra info) for an
+     * audit entry resolved by global EntryId. It must only render entries whose owning run is in the requested folder,
+     * otherwise any reader could read another folder's document audit history by guessing entry ids.
+     */
+    @Test
+    public void auditLogExtraInfoIsContainerScoped() throws Exception
+    {
+        UnitTestUtil.cleanupDatabase(DOC_GUID);
```
vagisha: Is this redundant with ` @After cleanupAuditLog()`?

test/src/org/labkey/test/tests/targetedms/passport/PassportTest.java
```diff
@@ -91,13 +93,20 @@ private void testNormalStuff()
...
+        waitForElementsAtLeast(Locator.tagWithClass("td", "feature-sequencevariant"), 3);
+        waitForElementsAtLeast(Locator.tagWithClass("td", "feature-glycosylationsite"), 3);
+        waitForElementsAtLeast(Locator.tagWithClass("td", "feature-helix"), 3);
+        waitForElementsAtLeast(Locator.tagWithClass("td", "feature-turn"), 3);
     }
 
+    public void waitForElementsAtLeast(final Locator loc, final int count)
+    {
+        waitFor(() -> loc.findElements(getDriver()).size() >= count, WAIT_FOR_JAVASCRIPT);
+        assertTrue("Element not present at least the expected number of times", loc.findElements(getDriver()).size() >= count);
```
vagisha, line 106: `loc.findElements(getDriver()).size()` is getting called twice.

vagisha (approved):
Thank you!

src/org/labkey/targetedms/TargetedMSContainerScopingTest.java
labkey-jeckels: Largely, yes. Better to mark that method as both `Before` and `After` and remove this call. Done.

test/src/org/labkey/test/tests/targetedms/passport/PassportTest.java
labkey-jeckels: True, but one is inside the lambda. We could reuse the result but I think it's a bit cleaner (and reasonable) to check again outside the lambda.

src/org/labkey/targetedms/TargetedMSContainerScopingTest.java
labkey-jeckels: Thanks. Fixed.

labkey-jeckels:
@DariaBod FYI, here's the fix for PassportTest. Merged to 25.7 earlier this morning.

PR #1257, opened 2026-07-07
Move TargetedMS QC/chart UI off ExtJS to plain JS + HTML

ankurjuneja:
## Rationale
We are gradually removing our dependency on ExtJS. ExtJS is old, heavy, and something we want to stop using. Each of these screens used ExtJS only as a wrapper around plain data or existing non-ExtJS chart code, so we can swap in simple HTML, plain JavaScript, and jQuery without changing what the user sees. This makes the code lighter, easier to read, and easier to maintain. 

## Related Pull Requests
- 

## Changes
- Trace metric dialog (resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js)
- Pareto plot (webapp/TargetedMS/js/ParetoPlotPanel.js)
- Calibration curve (webapp/TargetedMS/js/CalibrationCurve.js)
- Summary charts (src/org/labkey/targetedms/view/summaryChartsView.jsp)

### DIFF
--- a/resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
+++ b/resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
@@ -4,518 +4,263 @@
  * Licensed under the Apache License, Version 2.0: http://www.apache.org/licenses/LICENSE-2.0
  */
 
-Ext4.define('Panorama.Window.AddTraceMetricWindow', {
-    extend: 'Ext.window.Window',
+(function($) {
+    window.Panorama = window.Panorama || {};
+    window.Panorama.Window = window.Panorama.Window || {};
 
-    modal: true,
-    closeAction: 'destroy',
-    bodyStyle: 'padding: 10px;',
-    autoScroll: true,
-    border: false,
-    update: 'update',
-    insert: 'insert',
-    timeValueOptions:  Ext4.create('Ext.data.Store', {
-        fields: ['value'],
-        data : [
-            { "value":"First"},
-            { "value":"Last"},
-            { "value":"Max"},
-            { "value":"Min"}
-        ]
-    }),
+    const DIALOG_ID = 'lk-trace-metric-dialog';
+    const TIME_VALUE_OPTIONS = ['First', 'Last', 'Max', 'Min'];
+    let _config = null;
 
-    initComponent: function() {
-        var title = this.operation === this.insert ? 'Add New Trace Metric' : 'Edit Trace Metric';
-        this.setTitle(title);
-        this.height = Ext4.max([Ext4.getBody().getHeight() * 0.3, 250]);
-        this.width = Ext4.max([Ext4.getBody().getWidth() * 0.3, 600]);
-        this.items = this.getItems();
-        this.dockedItems= [{
-            xtype: 'toolbar',
-            dock: 'bottom',
-            ui: 'footer',
-            items: this.getButtons()
-        }]
+    function closeDialog() {
+        $('#' + DIALOG_ID).remove();
+    }
 
-        this.callParent();
+    function showError(msg) {
+        $('#lk-trace-metric-error').text(msg).show();
+    }
 
-        this.timeValue = true;
-        this.traceValue = false;
+    function clearErrors() {
+        $('#lk-trace-metric-error').hide().text('');
+        $('#lk-trace-metric-name, #lk-trace-use-trace, #lk-trace-ylabel, #lk-trace-time-option, #lk-trace-min-time, #lk-trace-max-time, #lk-trace-value')
+            .css('border-color', '');
+    }
 
-    },
+    function markInvalid($field) {
+        $field.css('border-color', 'red');
+    }
 
-    getItems: function() {
-        return [
-            this.getMetricNameField(),
-            this.getTracesCombo(),
-            this.getYAxisLabelField(),
-            this.getTraceValueRadioGroup(),
-            this.getQueryError()
-        ];
-    },
+    function getMode() {
+        return $('input[name="metricValue"]:checked').val(); // 'timeValue' or 'traceValue'
+    }
 
-    getButtons: function () {
-        var buttons = [];
+    // Enable only the fields belonging to the selected mode
+    function refreshMode() {
+        const isTime = getMode() === 'timeValue';
+        $('#lk-trace-time-option, #lk-trace-min-time, #lk-trace-max-time').prop('disabled', !isTime);
+        $('#lk-trace-value').prop('disabled', isTime);
+    }
 
-        buttons.push(this.getCancelButton());
-        buttons.push('->'); // to push remaining buttons to the right
-        if (this.operation === this.update) {
-            buttons.push(this.getDeleteButton());
+    function isNonNegativeNumber(val) {
+        return val !== '' && val !== null && !isNaN(val) && Number(val) >= 0;
+    }
+
+    function validate() {
+        clearErrors();
+        let isValid = true;
+
+        if (!$('#lk-trace-metric-name').val().trim()) {
+            markInvalid($('#lk-trace-metric-name'));
+            isValid = false;
         }
-        buttons.push(this.getSaveButton());
-        return buttons;
-    },
+        if (!$('#lk-trace-use-trace').val()) {
+            markInvalid($('#lk-trace-use-trace'));
+            isValid = false;
+        }
+        if (!$('#lk-trace-ylabel').val().trim()) {
+            markInvalid($('#lk-trace-ylabel'));
+            isValid = false;
+        }
 
-    getMetricNameField: function() {
-        if (!this.metricNameField) {
-            this.metricNameField = Ext4.create('Ext.form.field.Text', {
-                fieldLabel: 'Metric Name',
-                labelWidth: 150,
-                width: 570,
-                name: 'metricName'
-            });
-
-            if (this.operation === this.update) {
-                this.metricNameField.setValue(this.metric.name);
+        if (getMode() === 'timeValue') {
+            if (!$('#lk-trace-time-option').val()) {
+                markInvalid($('#lk-trace-time-option'));
+                isValid = false;
+            }
+            if (!isNonNegativeNumber($('#lk-trace-min-time').val())) {
+                markInvalid($('#lk-trace-min-time'));
+                isValid = false;
+            }
+            if (!isNonNegativeNumber($('#lk-trace-max-time').val())) {
+                markInvalid($('#lk-trace-max-time'));
+                isValid = false;
             }
         }
-
-        return this.metricNameField;
-    },
-
-    getTracesCombo: function () {
-      if (!this.tracesCombo) {
-          var config = {
-              fieldLabel: 'Use Trace',
-              name: 'useTrace',
-              labelWidth: 150,
-              width: 570
-          }
-          if (!this.tracesPresent) {
-              config.emptyText = 'No trace can be found';
-          }
-          else {
-              config.store = this.getTracesComboStore();
-              config.valueField = 'TextId';
-              config.displayField = 'TextId';
-          }
-          this.tracesCombo = Ext4.create('Ext.form.field.ComboBox', config);
-
-          if (this.operation === this.update) {
-              this.tracesCombo.setValue(this.metric.TraceName);
-              this.tracesCombo.bindStore(this.getTracesComboStore());
-          }
-      }
-      return this.tracesCombo;
-    },
-
-    getTracesComboStore: function () {
-        return Ext4.create('Ext.data.Store', {
-            fields: ['TextId'],
-            sorters: [{property: 'TextId'}],
-            data: this.traces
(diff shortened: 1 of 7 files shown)


src/org/labkey/targetedms/view/calibrationCurve.jsp
```diff
@@ -30,7 +30,7 @@
     @Override
     public void addClientDependencies(ClientDependencies dependencies)
     {
-        dependencies.add("Ext4");
+        dependencies.add("Ext4"); // LABKEY.vis (vis/vis) still uses Ext.isArray for log-scale axes
```
cnathe: @ankurjuneja this is the PR I was thinking of when I tagged you for the changes here https://github.com/LabKey/platform/pull/7865. Not sure if that changes this dependency at all, but wanted to call it out.

ankurjuneja: Thanks @cnathe, I will verify (and remove) if we need the Ext4 dependency for calibration curve.

resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
```diff
@@ -4,518 +4,263 @@
  * Licensed under the Apache License, Version 2.0: http://www.apache.org/licenses/LICENSE-2.0
  */
 
-Ext4.define('Panorama.Window.AddTraceMetricWindow', {
```
labkey-jeckels, line 7: `ConfigureMetricsUIPage` still has Ext locators.

resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
```diff
@@ -4,518 +4,263 @@
...
+    }
+
+    function buildDialogHtml() {
+        const op = _config.operation;
+        const metric = _config.metric || {};
+        const tracesPresent = !!_config.tracesPresent;
+        const title = op === 'insert' ? 'Add New Trace Metric' : 'Edit Trace Metric';
+
+        // In update mode, pick the mode based on the stored values; default to the time-value mode.
+        const isTraceValueMode = op === 'update' && metric.TraceValue > 0;
```
labkey-jeckels: The check on line 79 allows zeros when saving, but this treats them as a sentinel. Probably best to make line 79 ensure a positive value, which matches the old validation.

resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
```diff
@@ -4,518 +4,263 @@
...
-        }, this);
+        });
+    }
+
+    function buildTraceOptions(selectedTrace) {
+        const traces = (_config.traces || []).slice().sort(function(a, b) {
+            return String(a.TextId).localeCompare(String(b.TextId));
+        });
+        let html = '<option value="">-- Select trace --</option>';
+        traces.forEach(function(row) {
```
labkey-jeckels: Low priority, but if the previously saved value is no longer present, it won't be in the list. The old code always added the saved value. We could do the same here.

resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
```diff
@@ -4,518 +4,263 @@
...
+    function save() {
+        if (!validate()) return;
 
-                if (this.operation === this.update) {
-                    newMetric.id = this.metric.id;
-                }
+        const metricName = $('#lk-trace-metric-name').val().trim();
+        checkMetricNameExists(metricName, function(exists) {
+            if (exists) {
+                showError('A metric with the name "' + LABKEY.Utils.encodeHtml(metricName) + '" already exists. Please choose a different name.');
```
labkey-jeckels: `showError` uses `.text()` so this will be double-encoded. Remove the encoding here.

webapp/TargetedMS/js/CalibrationCurve.js
```diff
@@ -35,281 +37,288 @@ Ext4.define('LABKEY.targetedms.CalibrationCurve', {
...
+            ];
+        },
+
+        formatLegendValue: function(value, exp) {
+            if (value == null)
+                return 'NaN';
+            const rounded = Math.round(value * 100000) / 100000;
+            // Use scientific notation if the number is large or very small
+            if (exp && (rounded > 10000 || rounded < -10000 || (rounded > -0.00001 && rounded < 0.00001)))
+                rounded.toExponential(4)
```
labkey-jeckels:
```suggestion
                return rounded.toExponential(4)
```

src/org/labkey/targetedms/view/summaryChartsView.jsp
```diff
@@ -126,406 +138,262 @@
...
+                <td class="sc-label"><label for="sc-log">Log Scale</label></td>
+                <td><input type="checkbox" id="sc-log"></td>
+            </tr>
+            <tr>
+                <td class="sc-label"><label for="sc-width">Width</label></td>
+                <td><input type="number" id="sc-width" value="<%= bean.getInitialWidth() %>"></td>
+            </tr>
+            <tr>
+                <td class="sc-label"><label for="sc-height">Height</label></td>
+                <td><input type="number" id="sc-height" value="<%= bean.getInitialHeight() %>"></td>
```
labkey-jeckels:
```suggestion
                <td><input type="number" id="sc-height" value="<%= bean.getInitialHeight() %>" min="1" step="1"></td>
```

src/org/labkey/targetedms/view/summaryChartsView.jsp
```diff
@@ -126,406 +138,262 @@
...
+                <td class="sc-label"><label for="sc-cv">CV Values</label></td>
+                <td><input type="checkbox" id="sc-cv"></td>
+            </tr>
+            <tr>
+                <td class="sc-label"><label for="sc-log">Log Scale</label></td>
+                <td><input type="checkbox" id="sc-log"></td>
+            </tr>
+            <tr>
+                <td class="sc-label"><label for="sc-width">Width</label></td>
+                <td><input type="number" id="sc-width" value="<%= bean.getInitialWidth() %>"></td>
```
labkey-jeckels:
```suggestion
                <td><input type="number" id="sc-width" value="<%= bean.getInitialWidth() %>" min="1" step="1"></td>
```

src/org/labkey/targetedms/view/summaryChartsView.jsp
```diff
@@ -126,406 +138,262 @@
...
-            }
+            const areaElement = byId('peakAreasGraph');
+            LABKEY.targetedms.SVGChart.requestAndRenderSVG(peakAreaUrl, areaElement, null, byId('peakAreasGraphLabel'));
+            const timeElement = byId('retentionTimesGraph');
+            LABKEY.targetedms.SVGChart.requestAndRenderSVG(retentionTimesUrl, timeElement, null, byId('retentionTimesGraphLabel'));
+
+            areaElement.style.width = parseInt(widthInput.value) + 'px';
+            areaElement.style.height = parseInt(heightInput.value) + 'px';
+            timeElement.style.width = parseInt(widthInput.value) + 'px';
+            timeElement.style.height = parseInt(heightInput.value) + 'px';
```
labkey-jeckels: Ideally we'd handle a NaN as `type="number"` isn't particularly strict.

resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
```diff
@@ -4,518 +4,263 @@
...
-                        if (response && response.exception) {
-                            errorMessage = response.exception;
-                        } else if (response && response.message) {
-                            errorMessage = response.message;
-                        }
-                        this.queryError.setText(errorMessage);
-                        this.queryError.setVisible(true);
-                    }
-                });
+            LABKEY.Query.saveRows({
```
labkey-jeckels, line 187: In my local testing, this can be sluggish to save. There's no feedback for the user while it's working, which led me to click on Save a second time, which then produced a key violation. This change presumably hasn't impacted the perf, but it would be nice to show a spinner and disable the buttons to avoid a double-submit.

src/org/labkey/targetedms/view/calibrationCurve.jsp
labkey-jeckels: Reminder to check on this.

src/org/labkey/targetedms/view/paretoPlot.jsp
```diff
@@ -27,11 +27,10 @@
     @Override
     public void addClientDependencies(ClientDependencies dependencies)
     {
-        dependencies.add("Ext4");
+        dependencies.add("Ext4"); // still needed by QCMetricConfigLoader.js
```
labkey-jeckels: Looks like it's just for JSON parsing. Can we migrate that?

resources/web/PanoramaPremium/window/AddNewAnnotationMetricWindow.js
```diff
@@ -13,9 +13,28 @@
     let _allAnnotations = [];
 
     function closeDialog() {
+        $(document).off('keydown.lkAnnotationMetric');
         $('#' + DIALOG_ID).remove();
     }
 
+    // Ext.window.Window used to keep Tab inside the modal; do the same by hand
```
labkey-jeckels: No need to retain info on old implementation

```suggestion
    // Keep Tab inside the modal
```

resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
```diff
@@ -4,518 +4,336 @@
...
+
+    function closeDialog() {
+        $(document).off('keydown.lkTraceMetric');
+        $('#' + DIALOG_ID).remove();
+    }
 
-            if(this.operation === this.update) {
-                this.timeValueOptionField.setValue(this.metric.TimeValueOption);
-            }
+    // Ext.window.Window used to keep Tab inside the modal; do the same by hand
```
labkey-jeckels:
```suggestion
    // Keep Tab inside the modal
```

src/org/labkey/targetedms/view/summaryChartsView.jsp
```diff
@@ -98,6 +99,17 @@
     peakAreaUrl.addParameter("chartHeight", bean.getInitialHeight());
     retentionTimesUrl.addParameter("chartWidth", bean.getInitialWidth());
     retentionTimesUrl.addParameter("chartHeight", bean.getInitialHeight());
+
+    // The ExtJS stores prepended "All"/"None" entries, so the original visibility checks were
+    // count-based (e.g. store.count() > 2). Translate those to list-size checks here.
```
labkey-jeckels: Is this important post-migration?

webapp/TargetedMS/js/CalibrationCurve.js
```diff
@@ -5,27 +5,29 @@
  */
 /**
  * Created by Marty on 3/16/2017.
+ *
+ * Plain JS/HTML implementation (no ExtJS). Renders the calibration curve plot (LABKEY.vis)
```
labkey-jeckels:
```suggestion
 * Plain JS/HTML implementation. Renders the calibration curve plot (LABKEY.vis)
```

resources/web/PanoramaPremium/window/AddNewTraceMetricWindow.js
labkey-jeckels: Further/related concern:

Issue: const isTraceValueMode = op === 'update' && metric.TraceValue > 0; picks the radio button by looking at TraceValue alone, but the server resolves the opposite way — TargetedMSManager line ~2910 checks timeValueOption first and only falls through to traceValue when it is null.                                                                                                                                                                                  

Why it matters: Every trace metric saved by the old dialog can have both configurations populated, because the old saveNewMetric wrote only the chosen mode's columns and never cleared the other. Opening such a row now shows the trace-value radio selected with a trace threshold filled in, while the plot on the dashboard is still being computed from TimeValueOption/MinTimeValue/MaxTimeValue. That is the same class of "UI shows one thing, server does another" bug the rest of this change sets out to eliminate.                                                                                                    

Suggestion: const isTraceValueMode = op === 'update' && !metric.TimeValueOption && metric.TraceValue > 0;

src/org/labkey/targetedms/view/summaryChartsView.jsp
```diff
@@ -126,406 +138,273 @@
...
+                        <option value="All">All</option>
+                        <option value="Retention Time">Retention Time</option>
+                        <option value="FWHM">FWHM</option>
+                        <option value="FWB">FWB</option>
+                    </select>
+                </td>
+            </tr>
+            <tr>
+                <td></td>
+                <td><span class="valuelabel">Value only effects Retention Times chart.</span></td>
```
labkey-jeckels:
```suggestion
                <td><span class="valuelabel">Value only affects Retention Times chart.</span></td>
```

labkey-jeckels (approved):
Minor suggestions but not need for further review unless you want it.

PR #1264, opened 2026-07-22
Use centrally configured SchemaFactory

labkey-jeckels:
## Rationale
We should configure `SchemaFactory` centrally like other XML objects so that we get the right settings consistently.

## Related Pull Requests
- https://github.com/LabKey/platform/pull/7875

## Changes
- Use utility/factory method to get a configured `SchemaFactory` with external references disabled


labkey-adam:
@labkey-jeckels looks like some unhappy tests here. https://stackoverflow.com/questions/53299280/java-and-xerces-cant-find-property-xmlconstants-access-external-dtd

labkey-jeckels:
> @labkey-jeckels looks like some unhappy tests here. https://stackoverflow.com/questions/53299280/java-and-xerces-cant-find-property-xmlconstants-access-external-dtd

Fixed. The test failures in the recent run are now unrelated. I'll queue another run just to confirm.

labkey-adam (approved):

PR #1266, opened 2026-07-27
Fix instrument scheduler datetime-local fields showing midnight on Firefox

ankurjuneja:
## Rationale
The instrument reservation modal's start/end datetime-local fields were rendered with DateFormat.format.date(date, container.dateTimeFormat), a display format with a space separator. Firefox strictly requires the W3C wire format yyyy-MM-ddTHH:mm and blanks any other value, so both fields collapsed to midnight and the save was rejected with "StartTime must be before EndTime" — reproducible on Firefox in timezones like Honolulu.

## Related Pull Requests
- https://github.com/LabKey/platform/pull/7883

## Changes
- scheduleUtils.js: add shared toDateTimeLocalValue
- scheduler.js: populate the datetime-local inputs via ScheduleUtils.toDateTimeLocalValue
- instrumentCalendar.jsp: replace open-coded zero-pad date formatting with ScheduleUtils.toDateValue
- InstrumentSchedulingTest: assert the fields default to T08:00/T17:00 so a regression that zeroes or shifts the time fails on any agen

## Tasks 📍
- [x] Claude Code Review
- ~Manual Testing~
- [x] Test Automation

### DIFF
--- a/test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
+++ b/test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
@@ -200,6 +200,9 @@ public class InstrumentSchedulingTest extends TargetedMSTest implements Postgres
 
         assertProjectEventCounts(2, 0);
 
+        // The event chip time range is rendered via DateFormat (the patched parseTime path); verify it shows the real 8AM-5PM range, not a zeroed 00:00 - 00:00 as happened in colon-labelled timezones.
+        assertEquals("Event chip should show the 8AM-5PM time range", "08:00 - 17:00", getText(Locator.css(".activeProjectEvent .event-date")));
+
         doAndWaitForPageToLoad(() -> selectOptionByText(PROJECT_DROP_DOWN, PROJECT_2));
 
         scheduleInstrument(yearMonth + "-04", false);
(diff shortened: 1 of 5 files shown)


webapp/TargetedMS/js/scheduler.js
labkey-jeckels: Issue: Pre-existing, but in the function this diff modifies: #setup-cost, #instrument-fee, #total-cost, and #schedule-cost-error do not exist in scheduleInstrument.html (nor anywhere in the module — only #preview-* and #schedule-save-error do). So the four .val() calls in the success handler are no-ops, and the inner selectRows failure handler writes to nothing — a rate-type lookup failure produces no user-visible message at all.                 
Why it matters: The change's stated goal is that the cost log never shows garbage. It now bails silently on bad input and still fails silently on a query error, so a user who sees no cost row has no way to tell which happened.     
Suggestion: Route the failure text to #schedule-save-error (which exists) and delete the three dead .val() calls. While in there, Math.abs() on line 602 is now redundant — the new guard already establishes end > start.

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
```diff
@@ -197,6 +200,9 @@ public void testSchedule() throws IOException, CommandException
 
         assertProjectEventCounts(2, 0);
 
+        // The event chip time range is rendered via DateFormat (the patched parseTime path); verify it shows the real 8AM-5PM range, not a zeroed 00:00 - 00:00 as happened in colon-labelled timezones.
+        assertEquals("Event chip should show the 8AM-5PM time range", "08:00 - 17:00", getText(Locator.css(".activeProjectEvent .event-date")));
```
labkey-jeckels, line 204: Claude is concerned that this would fail if the server and browser are using different time zones. That's the Honolulu TZ config, so if that's passing, we should be fine.

labkey-jeckels (approved):

test/src/org/labkey/test/tests/targetedms/InstrumentSchedulingTest.java
ankurjuneja: that is the exact config this PR is fixing for TC - it passes now on TC.

